Issue #15, opened 2024-10-14
Implement utility functions for SR quantities.


dtabell:
Place functions for computing beta, gamma, Brho, _etc._ in `utils.jl`.

dtabell:
@DavidSagan and @mattsignorelli
It turns out (I should have remembered this) to be much better to
compute the various relativistic quantities starting from either the
kinetic energy, $E_\text{kin}$, or $\beta\gamma$. So the `beam` struct should store the
reference value of one of those quantities instead of $\beta_0$. I have a
slight preference for storing $(\beta\gamma)_0$—it is dimensionless—but I am
comfortable with either. Thoughts?

dtabell:
On the other hand, storing the kinetic energy would likely make it
easier for people to debug their simulations.

mattsignorelli:
My vote would be for the dimensionless quantity, so that all unitful quantities are contained in the `species` field

dtabell:
I implemented functions to compute the following:
- [x] $\gamma$
- [x] $\gamma - 1$
- [x] $\beta$
- [x] $pc$
- [x] $E_\text{kin}$
- [x] $E_\text{tot}$
- [x] $B\rho$

One important difference from what we discussed: I made the arguments
$E_\text{rest}$ and $\beta\gamma$. In other words, I decided we should pass in $E_\text{rest}$ instead of a
`Species`. This choice of arguments will speed computations that require
computing one of the above quantities for an entire beam.

DavidSagan:
@dtabell I have functions for computing these in AcceleratorLattice. We should combine and put in SimUtils
 or perhaps AtomicAndPhysicalConstants

mattsignorelli:
I've modified the `beta_0` field to `beta_gamma_0` field in `Beam`

dtabell:
I think we can close this issue.

PR #17, opened 2024-10-15
Implement functions for SRelativistic quantities.

dtabell:
This PR provides implementations of functions to compute accurate values
for a range of relativistic parameters and quantities — $\gamma$, $\gamma - 1$, $\beta$, $pc$, $E_\text{kin}$, $E_\text{tot}$,
and $B\rho$ — given values for the rest energy, $mc^2$, and $\beta\gamma$.

Items to address / comment on:
- [ ] function names
- [ ] function arguments — use E_\text{kin} instead of $\beta\gamma$?
- [ ] docstrings

The definitions given here will require corresponding changes to the tracking
routines. Should those changes be made part of this PR?

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,3 +1,125 @@
-function pc_ref(species::Species, beta_0)
-  return "hi!"
+"""
+utilities.jl
+"""
+
+
+"""
+    sr_gamma(beta_gamma)
+
+For a particle with relativistic parameter ``\\beta\\gamma``,
+compute the relativistic Lorentz factor ``\\gamma``.
+"""
+function sr_gamma(beta_gamma)
+  return hypot(1, beta_gamma)
 end
+
+
+"""
+    sr_gamma_m1(beta_gamma)
+
+For a particle with relativistic parameter ``\\beta\\gamma``,
+compute the relativistic Lorentz factor minus one, ``\\gamma - 1``.
+"""
+function sr_gamma_m1(beta_gamma)
+  return beta_gamma^2 / (hypot(1, beta_gamma) + 1)
+end
+
+
+"""
+    sr_beta(beta_gamma)
+
+For a particle with relativistic parameter ``\\beta\\gamma``,
+compute the normalized velocity ``\\beta = v / c``.
+"""
+function sr_beta(beta_gamma)
+  return beta_gamma / hypot(1, beta_gamma)
+end
+
+
+"""
+    sr_pc(e_rest, beta_gamma)
+
+For a particle with a given rest energy and relativistic parameter
+``\\beta\\gamma``, compute the energy-equivalent momentum, ``pc``.
+"""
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -1,3 +1,125 @@
-function pc_ref(species::Species, beta_0)
-  return "hi!"
-end
\ No newline at end of file
+"""
+utilities.jl
+"""
```
mattsignorelli:
```suggestion
#=

Utility functions for tracking.

=#
```

mattsignorelli:
Why is it simpler to use beta*gamma instead of just beta or gamma?

dtabell:
> Why is it simpler to use beta*gamma instead of just beta or gamma?

@mattsignorelli Numerical stability: Given $\beta$, the computation of $\gamma$ for ultra-relativistic
particles loses accuracy. And given $\gamma$, the computation of $\beta$ for very-low-energy particles
loses accuracy. You can't avoid subtraction. On the other hand given $\beta\gamma$, or $E_\text{kin} = mc^2\beta\gamma$,
we have the very useful identity $\gamma^2 = 1 + (\beta\gamma)^2$, with which one can — regardless of energy
— compute accurate values for the other relativistic parameters and quantities .

mattsignorelli:
Sounds good! Because of the discussion with @DavidSagan  #15 , I would say we can merge this for now and _not_ export the symbols, just so we can get tracking going, and then later decide where to let them live.

src/utils.jl
```diff
@@ -1,3 +1,125 @@
...
+
+
+
+export sr_gamma,
+       sr_gamma_m1,
+       sr_beta,
+       sr_pc,
+       sr_ekin,
+       sr_etot,
+       brho
```
mattsignorelli:
```suggestion
```

dtabell:
> @dtabell I have functions for computing these in `AcceleratorLattice`. We
should combine and put in `SimUtils` or perhaps `AtomicAndPhysicalConstants`.

@DavidSagan I'm puzzled as to why you need them in AcceleratorLattice. I did take
a look at your functions, and they are admirably flexible. But I rather like the idea of
forcing the user to give me $E_\text{kin}$ or $\beta\gamma$, because I know those allow me to compute
accurate value for everything else.

As to location, I think those functions belong in `BeamTracking` or, perhaps, `SimUtils`.
If `AcceleratorLattice` really needs them, then I favor `SimUtils`.

dtabell:
@mattsignorelli I implemented both of your suggested changes.
(Did it by hand because I did not notice the `Commit suggestion` buttons above.)
I'm leaving in the file the commented-out functions that take $E_\text{kin}$ instead of $\beta\gamma$,
just in case we later choose to switch. We can delete them once we're certain.

mattsignorelli (approved):

PR #22, opened 2024-10-25
Linear drift quad testing + single Beam struct + particle struct

sophiayang627:


mattsignorelli:
Thanks for all your work! I think it would be worth making your `particle` function instead be a constructor for some struct called `Particle`, since it could be very useful in the general case. Could you create a `Particle` struct with the same layout as `Beam`, but instead of `Coords` it just contains a `Coord` vector which you can define as

```julia
struct Coord{T} <: FieldVector{6, T} # Just like Coords but only 1 Coord 
  x::T
  px::T
  y::T
  py::T
  z::T
  pz::T
end
```

PR #37, opened 2025-01-14
4 mk drift

dtabell:
Updated the drift and quadrupole integrators.
Added tests for both integrators.


dtabell:
So ... when I use a relative tolerance of 5e-13, all tests pass on my machine.
But looking at the [CI Julia 1.11 test results](https://github.com/bmad-sim/BeamTracking.jl/actions/runs/12837304657/job/35800702093?pr=37), I see some tests failing with relative errors of order unity!
Extremely puzzling.

PR #38, opened 2025-02-15
Thin lense rf cavity

dohter:
Adding an RF cavity, thin lens, may need work on voltage units, please check


mattsignorelli:
closing stale PR

PR #42, opened 2025-04-06
added combined function +testings for it

sophiayang627:
This should be ready to merge


mattsignorelli:
It looks like the test for the combined function is failing?

sophiayang627:
> It looks like the test for the combined function is failing?

It looks like it not recognizing the tracking functions for combined function elements. I'll take a look and fix it. Not so sure why it happens :(

sophiayang627:
I've resolved the problem with edge tracking even with finite dg and the tests all runs successfully.

mattsignorelli:
@sophiayang627 Do you have some time today, maybe after the group meeting? I'm about to push a working example of tracking the EIC-ESR and I want to go through with you more in-depth the code, and how you can merge your combined function + finite "dg" with the new framework

sophiayang627:
> @sophiayang627 Do you have some time today, maybe after the group meeting? I'm about to push a working example of tracking the EIC-ESR and I want to go through with you more in-depth the code, and how you can merge your combined function + finite "dg" with the new framework

I can come over to the lab sometime before the group meeting, possibly around 12:00 if that works for you. I have class right after the meeting, so earlier would be better.

mattsignorelli:
I can't do before the meeting. When does your class end? Are you free some time after that?

sophiayang627:
My schedule is a bit tight today as I have another meeting right after my afternoon class…but I could come in tomorrow after 11am or 4:30 pm if that works for you!

mattsignorelli:
Does 11:30am today work for you? I finished a bit earlier my tasks for this morning but I will need to bow out at 12 for lunch until the 12:30 meeting

mattsignorelli:
outdated per #54

PR #46, opened 2025-04-08
various ideas for chains

mattsignorelli:
trying out my "chain" solution in different ways. Here's what I've found.
Firstly, it turns out that iterating over a Tuple is type unstable if the arguments are all different. Specifically, if I have 
t = (1, "hi", 3.0f0) for example, then doing t[1] is type stable, but doing t[i] in a for loop is not. This was annoying to find out, because that means my original idea of having a tuple of each function call and its arguments is very type unstable. I tested this solution and indeed, it was significantly slower.
The internet showed me that there are two ways to handle this: one is to recursively go element-by-element in the tuple. This I found to be type stable, but allocations started happening at 32 elements in the tuple anyways, not exactly sure why. Also, the autovectorization of course did not happen with a recursive call inside the parallel for loop, so I didn't like  this solution. It was slower than my launch! inside every element method that I showed last week.
The second option is to use `@generated` functions, which read the Tuple at compile time, fully unroll the tuple, and basically cut and paste the tracking routines into the for loop over each particle. This I was hoping would work, as we would basically be compiling a separate tracking function for every lattice, but unfortunately the compile time grew unbearably long with just 1000 lattice elements. Furthermore, it was still significantly slower than my initial approach. (maybe this would be faster on a GPU though? I don't know).
Then, I tried implementing the chain as a Vector{KernelCall} instead of a tuple, where KernelCall is a type storing the function and arguments. Of course there is a type instability here because each function is its own type and the argument list can be anything for each function. It too was almost always slower than my initial solution, though slightly faster when 16 cores on this Mac M2 are used with manual SIMD.
Finally, I tried implementing the chain as an NTuple, where each tracking method and arguments are all the same. This actually did well but required manual SIMD. I think the solution will be some kind of universal track function with uniform arguments (types!) for each, 

```julia
track_universal!(i, v, bmultipoleparams, bendparams, alignmentparams, tracking_method_type::Type) 
  if tracking_method_type == Linear # NOTE: NOT Linear() but Linear so it is the type    
    track_universal_linear!(...) 
  elseif tracking_method_type == MatrixKick 
    track_universal_matrixkick(...) 
   ... 
   end
end
```
where for all elements, bmultipoleparams, bendparams, alignmentparams have the same type element-by-element. They will basically need to be arrays or fully type-stable structures.


mattsignorelli:
Producing a ntuple chain for my test lattice takes about 70ms, and tracking 100000 particles through around 20 ms including optimized multithreading. For reference, dynamic tracking of the same bunch + lattice is about 150ms including multithreading. So this solution is highly beneficial in the case of a static lattice, and even in the case when the lattice could be changing (required recomputation of the chain each turn).

@DavidSagan we should come up with a generic interface of all the arguments which all tracking methods implemented here should work with (this will be required for the GPU), e.g. misalignments, patch, how to specify multipole strengths (array?), etc. We can then allow people to either track a chain in the ntuple fashion, or track dynamically through each element as in my current implementation

mattsignorelli:
Really what it comes down to is for GPU tracking or "faster" tracking, we need a concrete element type definition, and move information over to run time. here is one idea

```julia
struct SuperParams{T}
  active::Bool    # true if the element has this parameter group and values in it, false if otherwise
  parameters::T  # if active == false then this will be said parameter group with all zeros
end

struct ConcreteElement{A<:Union{SuperParams,Nothing},B<:Union{SuperParams,Nothing},C<:Union{SuperParams,Nothing}}
   tracking_method::String
   bmultipole::A
   bend::B
   misalign::C
end

# ALL elements in the lattice must be the SAME type of `ConcreteElement{A,B,C}` exactly. If all elements do not have
# a specific parameter group, then that field can be `nothing`, but if a single one has it then the rest of them that do not 
# have it must have an "inactive" `SuperParams` field there.

...

function track_universal!(bunch, ele, work)
  if ele.tracking_method == "Linear"
    linear_universal!(bunch, ele, work)
  elseif ....
  ...
end

function linear_universal!(bunch, ele, work)
  if !isnothing(ele.misalign) || ele.misalign.isactive  # note that if it is nothing then that means entire lattice has no misalignments, so we compile it away statically
    misalign!(bunch, ele, work)
 ....
...
end

# now the entire lattice can be represented as an `NTuple` of `ConcreteElement{A,B,C}` (this is the chain), 
# and we can `launch!` on that 
```
And we will have to think about how to structure these parameters in superparameters. I think BMultipole for example could be a staticarray (up to specified order in entire lattice) or just a regular array for example

mattsignorelli:
I think an `ImmutableDict` for the bmultipole would be a good idea - the code would then still work completely fine with my regular dictionary in bmultipole too so we still have universally polymorphic cpu/gpu code

mattsignorelli:
OK so `Base.ImmutableDict` is NOT an isbits type, however `OrderedCollections.LittleDict`, when constructed with `SVector`s, is. Therefore, this will be the solution for `ImmutableBMultipoleParams`. The chain creation will do one pass over the lattice to get the max number of multipoles per element, and the highest-promoted multipole number type. The second pass will then create the immutable multipoles for each

Of course this will be a good solution for the static lattice, however for the case of ramping I would like chain creation to be fast enough that it still pays off for many particles (for the step ramp each turn). For optimization/TPSA, ALL number types must be TPSA/Dual and this is impractical on the GPU, so we stick with the current very-fast CPU implementation

mattsignorelli:
The plan is to convert the isbits lattice definition to an SoA layout so that struct padding is no longer an issue. Then I will keep the AoS "view" which can be created on the GPU and manipulated like a regular element in the kernel. I will also flatten the bits element fully as these paramter group separations can be very limiting ( e.g. no need to waste 3 floats on euler angles for misalignments if I only have translational offsets in my lattice)

mattsignorelli:
E multipolescan be the negative Int8 orders, and starting at typemin(Int8) I can add other things like RF voltage mappings. This can be good becaue e.g. now not all elements needs voltage and if only rf cavity has voltage then no memory used

PR #54, opened 2025-04-23
adding bend with dg

sophiayang627:


sophiayang627:
The linear_combined_matrices function should now handle the pure bend case correctly. I didn't remove linear_bend_matrices because it seems to be used in beamlinesExt.jl for testing.

src/modules/LinearTracking.jl
```diff
@@ -180,7 +180,7 @@ function linear_bend_matrices(K0, L, gamma_0, e1=nothing, e2=nothing)
   return mx, my, r56, d, t
 end
 
-function linear_combined_matrices(K0, g, K1, L, gamma_0, e1=nothing, e2=nothing)
+function linear_combined_matrices(K0, L, gamma_0; g=K0, K1=0.0, e1=nothing, e2=nothing)
```
mattsignorelli:
```suggestion
function linear_dipole_matrices(K0, L, gamma_0; g=nothing, K1=nothing, e1=nothing, e2=nothing)
```

Instead of `K1` defaulting to `0.0` (which is a `Float64`), it should default to `nothing` so that users may take advantage of compile-time optimizations if the bend does not have any `K1`. Also note that we would want to specify quantities like this as `0` instead of `0.0` because `0` is an `Int64`, and `promote_type(Int64,Float32)=Float32)`. So if users want to user lower precision floats, they can. 

Also because a kicker is `g=0` with `K0 != 0`, I suggest renaming this to `dipole` so it is more general than just a bend. 

Finally I suggest not defaulting `g=K0` for synonymous behavior with `Beamlines`. Instead we let it be `nothing` so that for kicker elements (which have `g=0`) users can gain compile-time optimizations

mattsignorelli: If you agree with my suggestions, then could you modify the source code to do similar `isnothing ... || .. == 0` for the `g` and the `K1`, like I do with the `e1` and `e2`?  Once that's in and the kernels have been added to `BeamlinesExt` then this will be ready to merge

sophiayang627: will do!

mattsignorelli:
Thanks! See the code coverage report and make sure these lines are covered: https://app.codecov.io/gh/bmad-sim/BeamTracking.jl/pull/54?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bmad-sim

mattsignorelli:
Also i see some repetitive computations (lines 249 and 250), could this be optimized? and I noticed that g != K0 does not affect the edge angle tracking. Is that correct?

sophiayang627:
> Also i see some repetitive computations (lines 249 and 250), could this be optimized? and I noticed that g != K0 does not affect the edge angle tracking. Is that correct?

Yup, the repetitive computation was a mistake—I’ll fix it. I initially thought that g !=K0 would affect the edge angle tracking, but it doesn’t, which is why it was incorrect before.

mattsignorelli:
> > Also i see some repetitive computations (lines 249 and 250), could this be optimized? and I noticed that g != K0 does not affect the edge angle tracking. Is that correct?
> 
> Yup, the repetitive computation was a mistake—I’ll fix it. I initially thought that g !=K0 would affect the edge angle tracking, but it doesn’t, which is why it was incorrect before.

@DavidSagan  is this correct? seems weird to me

mattsignorelli:
OK this is almost ready to merge, but I noticed that your orders of the multipoles (keys) are incorrect. Note that the keys in the dictionary (orders) are the orders of the cyclic group of the multipole, or equivalently the order of the scalar potential. So in fact:

| Order | Magnet |
| ------ | ---------- |
| 0 | Solenoid |
| 1 | Dipole |
| 2 | Quadrupole | 
| 3 | Sextupole |


Can you modify the code in Beamlines extension to reflect this properly?

sophiayang627:
Hi Matt, could you point out explicitly where the index was incorrect? I want to make sure I’m not misunderstanding.

The main modification I made was:

1. On line 92, I allow combined bend elements with quadrupole strength (K1), but raise an error if the element contains higher-order multipoles.

2. Lines 98–103 are meant to check whether the element includes K1. If it does not, it falls back to the standard bend routine that requires g ≈ K0 (line 99) and sets K1 to nothing (line 101). Otherwise, if the element includes K1, K1 is stored for use in the combined case.

This logic is placed inside the dipole case (line 85: elseif haskey(bmultipoleparams.bdict, 1)), so I assumed I’m checking whether the element has quadrupole strength in addition to dipole strength.

Let me know where I went off with the indexing or interpretation. Thanks!

ext/BeamTrackingBeamlinesExt/linear.jl
```diff
@@ -89,16 +89,21 @@ function linear_universal!(
     if L == 0
       error("Thin bend not supported yet")
     end
-    if any(t -> t == 0 || t > 1, keys(bmultipoleparams.bdict))
-      error("Combined function bend tracking not implemented yet")
+    if any(t -> t == 0 || t > 2, keys(bmultipoleparams.bdict)) 
+      error("Linear tracking does not support combined bend tracking with order > 2")
```
mattsignorelli:
```suggestion
      error("Linear tracking does not support bend tracking including any other multipole except a quadrupole ")
```

ext/BeamTrackingBeamlinesExt/linear.jl
```diff
@@ -89,16 +89,21 @@ function linear_universal!(
...
     end
 
     K0 = get_thick_strength(bmultipoleparams.bdict[1], L, bunch.Brho_ref)
-
-    if !(K0 ≈ bendparams.g)
-      error("Linear tracking requires BendParams.g ≈ BMultipoleParams.K0")
+    if !haskey(bmultipoleparams.bdict, 2)
+      if !(K0 ≈ bendparams.g) 
+        error("Linear tracking requires BendParams.g ≈ BMultipoleParams.K0 for pure bend")
```
mattsignorelli: This is not true anymore now that you have implemented the code for `g != K0` right?

ext/BeamTrackingBeamlinesExt/linear.jl
```diff
@@ -89,16 +89,21 @@ function linear_universal!(
...
+      end
+      K1 = nothing
+    else
+      K1 = get_thick_strength(bmultipoleparams.bdict[2], L, bunch.Brho_ref)
```
mattsignorelli: Line 103 needs coverage in the tests

sophiayang627: Will do!

ext/BeamTrackingBeamlinesExt/linear.jl
sophiayang627: For the case of pure sbend (without K1 present), K0 will need to be equal to g to ensure accuracy. the dg implementation was only for combined cases.

mattsignorelli: But g != K0 is supported by the code you wrote. Just set K1 = 0. So you can have a check if `!(g \approx K0)` then use your combined code with K1 = 0, else use the "pure" sbend. These both should converge as g->K0 and K1->0

sophiayang627: I double checked with David and he says that Bmad uses the exact map when k1 is zero and does not use the same set of routine with K1 !=0. This might explain why the linear solutions does not converge with K1->0.
 
And I think in previous meetings we figured out that for a linear solution in pure sbend (that supports dg) we would have to linearize the exact solutions and that complicates things.

sophiayang627:
I tried a different testing with a smaller dg and the matrix converges with Bmad, so the discrepancy between the two routine in Bmad is probably due to paraxial approximation when dg is too large. So the code should be ready to merge now.

mattsignorelli:
OK! can you please just correct the branches in BeamlinesExt/linear.jl? E.g. we do not need g = K0 as you found. Also since you check for bendparam in every multipole order I would instead at the beginning check for bendparams, then if no bendparams the multipoles can be checked. Just to keep the code simple and clean and minimize repetitive checks for bendparams. Also the merge conflict will need to be resolved

sophiayang627:
I've updated BeamlinesExt/linear.jl to correct the branches and restructured the code to check for bendparam first. However, there's a method error involving linear_coast_uncoupled! on line 133 that I'm not sure how to resolve.

src/modules/LinearTracking.jl
```diff
@@ -149,32 +149,141 @@ function linear_solenoid_matrix(Ks, L)
...
+        z1  = -K0 * sxc
+        z11 = sgnx * abs(kx) * (L - cx * sxc) / 4
+        z12 = -sgnx * abs(kx) * sxc * sxc / 2
+      
+        mx = SA[cx  sxc; sgnx*abs(kx)*sxc  cx]
+        my = SA[cy  syc; sgny*abs(K1)*syc  cy]
+        r56 = (-dx_c * z1 - K0 * L * dx_c + L/gamma_0^2) 
+        d = SA[dx_c * (1 - cx), -sgnx * wx * (dx_c * wx * sxc), 0, 0]
+        t = SA[z1, z2, 0, 0]
+      else
```
eiad-hamwi, line 181: @sophiayang627
In the branch starting at line 206, consider the case `isnothing(g) && isnothing(K1) && K0 != 0`

This is precisely what a kicker magnet is: it has a dipole field (K0, K0L, B0, or B0L) but does not bend the coordinate system. However, with this control flow a kicker/steering magnet would be interpreted as a bend, because it is lumping together the different cases: `isnothing(g)` and `g ≈ K0`.

In other words, we should be distinguishing between the two cases:

1. `isnothing(g) || g == 0` 
2. `!isnothing(g) && g != 0`

You should also make these changes in `ext/BeamTrackingBeamlinesExt/linear.jl` because this can be caught early in the first branch (`isactive(bendparams)`).

If you don't want to implement the tracking for these steering magnets right now, just make the distinction in the code and throw an `error("Linear kickers/correctors/steering magnets not yet implemented")` then later we can implement it in another branch or pull request. I would be happy to do that myself.


Otherwise, you can find the relevant equations in Bmad manual section 26.13.

To address rotated non-bending dipoles, i.e. vertical correctors, just look at the parameter `bmultipoleparams.bdict[1].tilt0` which would be `nothing` or `0` for horizontal correctors, and `pi/2` for vertical correctors. @DavidSagan correct me if I'm wrong about the sign of `tilt0`.

DavidSagan: If there is a finite `K0` then `tilt0` should not be `nothing`.

mattsignorelli:
Hi @sophiayang627 

I will soon be merging a branch which unifies the unpacking step across all tracking methods into a set of unit, understandable functions for the tracking methods to override. I have also added comments specifying where you can inject your code into in the linear.jl file. It should be very easy for you to do once I get this merged. See the "unpack" branch.  I'll let you know when its merged

src/modules/LinearTracking.jl
eiad-hamwi: Thanks David.
@mattsignorelli I was thinking maybe instead of saying 
```
if isnothing(g) && K0 != 0
  g = K0
end
```
we could do:
```
if isnothing(K0) && g != 0
  K0 = g
end
```

That way people don't have to supply a `K0` as well as a `g` for a bend, but only a `g`.

sophiayang627: Thanks Eiad and David for the comments!

Currently, g defaults to nothing while K0 doesn’t have a default. So If we do:

```julia
if isnothing(K0) && g != 0
  K0 = g
end
```
Would it make more sense to set the parameter the other way around instead? Also, I’m curious what would be the key difference between the two approaches? Is one of them more likely to be explicitly set by the user?

Also from Friday's conversation I thought that  `isnothing(g)` would mean that `g = 0` instead of `g = K0`? Please let me know if I have misinterpreted this.

sophiayang627:
Got it!

src/modules/LinearTracking.jl
mattsignorelli: Hi Sophia, 

> Also from Friday's conversation I thought that isnothing(g) would mean that g = 0 instead of g = K0? Please let me know if I have misinterpreted this.

This is correct.

Wait until I merge unpack (happening in a few minutes now), and youll see how clear it should be with my comments :)

mattsignorelli:
OK @sophiayang627  !

Please pull the latest main branch and look in `ext/BeamTrackingBeamlinesExt/linear.jl`. There you will see I have commented your name and some details in all spots for you to add your code into. Note that here is "layer 2" of the package, so in each of these places you should just call some utility function which you add into the `src/modules/LinearTracking.jl` module (layer 1) to actually do the computations. The return should be a KernelCall. See the other functions (`thick_pure_bquadrupole`, `thin_pure_bquadrupole`, etc) in the linear.jl file for examples.

Note that in these function names, "pure" means only one single BMultipole, and "bend" means that there is a BendParams (g, e1, e2). In SciBmad the current convention is that `g` does not have ANY implication of any magnetic field - it is solely a curvature of the coordinate system through the element.  This is unlike current Bmad, which I describe in my descriptions in linear.jl for you. 

This convention may change in the future but for now let's just keep doing it this way.

Let me know if you have any questions

ext/BeamTrackingBeamlinesExt/linear.jl
```diff
@@ -45,9 +45,10 @@ end
 @inline function drift(tm::Linear, bunch, L)
   gamma_0 = calc_gamma(bunch.species, bunch.Brho_ref)
   return KernelCall(LinearTracking.linear_drift!, (L, L/gamma_0^2))
+
```
mattsignorelli, line 48:
```suggestion
```
no need for line here

mattsignorelli:
Sorry I will wait until this is ready, just lmk when this is ready for my review

sophiayang627:
I have filled in the BeamlineExt so it is ready for review! I have also fixed the utility function` linear_dipole_matrices` so now `g = Nothing` refers to the right case (`g = 0`). To make g and K1 type stable I've defined them as `::Union{Nothing, Real}` that defaults to nothing. 

I'll put in more tests to make sure that all lines are covered.

ext/BeamTrackingBeamlinesExt/linear.jl
```diff
@@ -58,15 +58,22 @@ end
   # Sophia: this a thick corrector coil
   # In Fortran Bmad is g == 0 but dg != 0. 
   # In SciBmad this is g == 0 but K0 != 0.
-  error("Undefined for tracking method $tm")
+  gamma_0 = calc_gamma(bunch.species, bunch.Brho_ref)
+  K0 = get_thick_strength(bm1, L, bunch.Brho_ref)
+  mx, my, r56, d, t = LinearTracking.linear_dipole_matrices(K0, L, gamma_0; e1=bendparams.e1, e2=bendparams.e2)
```
mattsignorelli: There is no `bendparams` in this scope because this is a corrector coil (so no coordinate system curvature). This code would throw an error if reached

src/modules/LinearTracking.jl
```diff
@@ -125,32 +125,151 @@ function linear_solenoid_matrix(Ks, L)
...
-    t = SA[t[1]+me1*t[2], t[2], 0, 0]
-  end
 
-  if !isnothing(e2) && e2 != 0
-    me2 = K0*tan(e2)
-    mx = SA[1 0; me2  1]*mx
-    my = SA[1 0; -me2 1]*my
-    d = SA[d[1], me2*d[1]+d[2], 0, 0]
-  end
+function linear_dipole_matrices(K0, L, gamma_0; g::Union{Nothing, Real}=nothing, K1::Union{Nothing, Real}=nothing, e1=nothing, e2=nothing)
```
mattsignorelli: We cannot restrict any number type to `Real`, because e.g. `TPS <: Number` and `!(TPS <: Real)`

mattsignorelli:
```suggestion
function linear_dipole_matrices(g, e1, e2, K0, K1, gamma_0, L)
```
Can we just use this for this function definition instead? I think that will simplify things a lot. Then the code will just check if they are zero and do the corresponding operations.

The idea of using `nothing` was to get some compile-time optimizations, but it is becoming too complicated and with any type instability we can get big performance hits

mattsignorelli (changes requested):
Nice work! This is almost ready. Just some minor corrections suggested. Also you will need to add tests to cover the uncovered lines: https://app.codecov.io/gh/bmad-sim/BeamTracking.jl/pull/54?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bmad-simhttps://app.codecov.io/gh/bmad-sim/BeamTracking.jl/pull/54?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=bmad-sim

Once that's done then we can merge

src/modules/LinearTracking.jl
sophiayang627: I think this approach works as long as we ensure we don’t assign `nothing` to these values.

Would it perhaps be better to default `K0` and `K1` to `0` instead of `nothing`? That way, we wouldn’t need to specify them every time when they’re not used. Since `e1` and `e2` don’t have type instability issues, I think it makes sense to keep them as they are.

mattsignorelli: Let's let all of them default to `0`, and keep the checks `if __ \approx 0`

sophiayang627:
I have added tests to cover the uncovered lines

ext/BeamTrackingBeamlinesExt/linear.jl
```diff
@@ -58,10 +58,7 @@ end
   # Sophia: this a thick corrector coil
   # In Fortran Bmad is g == 0 but dg != 0. 
   # In SciBmad this is g == 0 but K0 != 0.
-  gamma_0 = calc_gamma(bunch.species, bunch.Brho_ref)
-  K0 = get_thick_strength(bm1, L, bunch.Brho_ref)
-  mx, my, r56, d, t = LinearTracking.linear_dipole_matrices(K0, L, gamma_0; e1=bendparams.e1, e2=bendparams.e2)
-  return KernelCall(LinearTracking.linear_coast_uncoupled!, (mx, my, r56, d, t))
+  error("Undefined for tracking method $tm")
```
mattsignorelli, line 62:
```suggestion
  # Singularity for `g==0`
  error("Undefined for tracking method $tm")
```

We should add a comment as to why this isn't supported

src/modules/LinearTracking.jl
```diff
@@ -126,10 +126,10 @@ end
 
 
 
-function linear_dipole_matrices(K0, L, gamma_0; g::Union{Nothing, Real}=nothing, K1::Union{Nothing, Real}=nothing, e1=nothing, e2=nothing)
+function linear_dipole_matrices(K0, L, gamma_0; g=0, K1=0, e1=nothing, e2=nothing)
```
mattsignorelli: `e1` and `e2` should also default to `0`

mattsignorelli: Keyword arguments here also don't make too much sense, I would just use the scheme I suggested in my previous review

mattsignorelli (changes requested):

ext/BeamTrackingBeamlinesExt/linear.jl
```diff
@@ -58,6 +58,7 @@ end
   # Sophia: this a thick corrector coil
   # In Fortran Bmad is g == 0 but dg != 0. 
   # In SciBmad this is g == 0 but K0 != 0.
+  # Singularity for simutaneous `g==0' and 'k1 = 0'
```
mattsignorelli:
```suggestion
  # Singularity for g==0
```

mattsignorelli (approved):

mattsignorelli:
Thanks for all of your work on this!

PR #79, opened 2025-05-08
Patch tracking

eiad-hamwi:
- Added exact patch tracking function `patch!` which transforms coordinate system and drifts element to the face of the new coordinate system. 
- Added utility functions `w_matrix` and `w_inv_matrix` for generating rotation matrices.
- Added tests which compare jacobian of `patch!` with imported Bmad 1st order matrix.

Note:
- `dt` param does not influence `patch`, since reference time is not stored in particles.
- `L` param does not influence `patch!`, since drift-to-face distance depends on other params. 
- `L` param will still influence other tracking routines that care about length, such as multipole fields.

Future changes:
- Decide whether the `patch!` transform is applied _before_ or _after_ other tracking params that an element has, such as `BMultipole`. This could be a Boolean.

P.S. Not sure why the whole Manifest.toml was included in the pull request...


mattsignorelli:
Thanks for your work on this!

We should use the tools in `ReferenceFrameRotations.jl` for the `w_matrix` function. E.g. see https://juliaspace.github.io/ReferenceFrameRotations.jl/stable/lib/library/#ReferenceFrameRotations.angle_to_dcm-Tuple%7BNumber,%20Symbol%7D

mattsignorelli:
Oops sorry this is probably the correct function: https://juliaspace.github.io/ReferenceFrameRotations.jl/stable/lib/library/#ReferenceFrameRotations.angle_to_rot-Tuple{Number,%20Symbol}

eiad-hamwi:
My pleasure! Do you mean I should just `import ReferenceFramRotations: angle_to_rot`
Otherwise it seems a bit overkill to use an entire package just for one of its functions, no?

One nice thing about `angle_to_rot` is that we can use any convention for rotations that we wish, although it's probably still good to only use the SciBmad/Bmad/MAD default.

mattsignorelli:
`ReferenceFrameRotations` is relatively lightweight, and its features may be useful in many other contexts, so I do not see a problem with adding it as a dependency and including the whole package

eiad-hamwi:
It turns out there are always monomials in SciBmad which are missing from Bmad/PTC, but not the other way around. Some of the monomials are odd orders of `pz`, but not always. 
If I just compare the coefficients that are present in both SciBmad and Bmad/PTC to 10th order, each individual coefficient agrees to a tolerance of `1e-7`. This does get better at lower order.

I also checked that lower order terms in Bmad are always present in the higher-order Bmad maps, so that's not an issue.

DavidSagan:
@eiad-hamwi Are these monomials in SciBmad that are missing from PTC always at the maximum order?

eiad-hamwi:
@DavidSagan After some troubleshooting, it looks like there is only 1 new term that shows up at 9th order `pz^9`, and then 21 new terms at 10th order. 
At lower order than 9th, the only new terms that show up are machine precision errors around zero.

DavidSagan:
So if you set the calculation order to 8 everything agrees?

eiad-hamwi:
Yes, but I can only get agreement to 1e-8 using ‘|c1 - c2| < max( ε, ε
(|c1| + |c2|) )’ with all the terms.

eiad-hamwi:
Yes, but I can only get agreement to `1e-8` using `|c1 - c2| < max( ε, ε (|c1| + |c2|) )` with all the terms.

DavidSagan:
Thanks for the info. If you set the calculation order to 12 do the same terms disagree?
One way to help resolve this would be to use Mathematica to do the calculation.
PTC goes through extra calculational steps to convert forwards and backwards between Bmad phase space coords and PTC coords. This may be a factor but I don't know how.

eiad-hamwi:
The same terms and some more are missing at 12th order in Bmad/PTC.

I'm trying to resolve this with Mathematica, but the direct symbolic evaluation will absolutely die trying to evaluate this 9th order expansion in 6 variables. While it's doing this, I also wrote a simple C-compiled TPSA code in Mathematica, and will use it to evaluate the 9th order expansion, however it took 15s for 4th order and 75s for 5th order, so I don't have much hope for it either at that rate. I can see there's a reason we're using Laurent's CTPSA libraries for GTPSA, haha.

Where can I find the source code for PTC's patch element (which is where the Bmad coefficients come from, I presume)?

Personally, I am not too interested in disagreements between these two codes at 9th and 10th order. This is feeling quite  frivolous, and I'm afraid I'll probably move on to other SciBmad efforts rather than continue to invest time to resolve this disparity.

eiad-hamwi:
@DavidSagan @mattsignorelli actually, I take that back. I just remembered that the first nonzero term that's missing in Bmad/PTC is the _pure_ term `pz^9` in the monomials of `z`. This means I can just set all other coordinates to zero and evaluate the single variable expansion w.r.t `pz` to 9th order and above, which is orders of magnitude easier than the mixed terms.

Lo and behold, Mathematica's symbolic series expansion says that this coefficient should be `-0.0180682`, whereas SciBmad says `-0.0180681555...`, and in fact this term is missing from Bmad/PTC even if I calculate 10th- and 12th-order maps.

So, I would say that actually PTC has a bug. Case closed for me, yay.

DavidSagan:
@eiad-hamwi This issue is concerning in that it might be pointing to an underlying bug that we need to know about. Can you post the input parameters you are using?

eiad-hamwi:
@DavidSagan here is the Bmad lattice file I'm using:
```
beginning[p0c] = 10e6

parameter[geometry] = open
parameter[particle] = electron
parameter[taylor_order] = 9

p: patch, x_offset = 1, y_offset = 2, z_offset=3, y_pitch = 4, x_pitch = 5, tilt = 6

lat: line = (p)
use, lat
```

DavidSagan:
Using such large values for x_pitch and y_pitch may be problematical. Try values of 0.4 and 0.5. How does this compare?

eiad-hamwi:
@DavidSagan It seems like the `pz^9` term is still missing for small x_pitch, y_pitch, and tilt as well.

DavidSagan:
which of the 6 phase space output coords is this pz term?

DavidSagan:
And are the initial coordinates zero?

eiad-hamwi:
It is in the `z` coordinate, which should be at the very end of the outputted list from Bmad.
Yes, the initial coordinates are all zero to machine precision.

DavidSagan:
Now set everything except z_offset to zero and check. If there is still a problem then we can ask Etienne about this.

eiad-hamwi:
Yeah, for some reason even in this case with only nonzero `z_offset`, the pure `pz^n` coefficients are missing from `z` for all `n≥9`. Those terms should definitely be present because of the contribution to time-of-flight delay:
```
z_offset * (β/β0 - 1) 
= z_offset * (sqrt((pc)^2 + (mc^2)^2)/sqrt((pc*(1+pz))^2 + (mc^2)^2) - 1)
```

mattsignorelli:
I agree that we need to get to the bottom of this

DavidSagan:
Looks like a drift has the same problem.

eiad-hamwi:
Well, at least it's self-consistent 😅

DavidSagan:
@mattsignorelli Could you run PTC direct (not via Bmad) to check if this is a PTC problem or a Bmad/PTC interface problem?

mattsignorelli:
@DavidSagan  I don't know exactly how to run PTC, only FPP

DavidSagan:
Running 12th order for a patch (but not a drift) I see a small zeroth order term!
```
 Y :   -247.500000000000   0  0  0  5  0  7       12
 ---------------------------------------------------
 Py:      1.000000000000   0  0  0  1  0  0        1       0.000000000
 ---------------------------------------------------
 Z :      0.000000014701   0  0  0  0  0  0        0       0.000000000
 Z :      1.000000000000   0  0  0  0  1  0        1
 Z :      1.926231920007   0  0  0  0  0  1        1
 Z :     -1.000000000000   0  2  0  0  0  0        2
 Z :     -1.000000000000   0  0  0  2  0  0        2
 Z :     -0.106570843225   0  0  0  0  0  2        2
 Z :      2.000000000000   0  2  0  0  0  1        3
 Z :      2.000000000000   0  0  0  2  0  1        3
 Z :     -0.028972341078   0  0  0  0  0  3        3
 Z :     -0.750000000000   0  4  0  0  0  0        4
 Z :     -1.500000000000   0  2  0  2  0  0        4
 Z :     -0.750000000000   0  0  0  4  0  0        4
 Z :     -3.000000000000   0  2  0  0  0  2        4
```

eiad-hamwi:
With just a `z_offset`? That's not supposed to be there. 
On my version at least (latest conda install), I don't see a zeroth order term in `z` as 12th order with only `z_offset ≠ 0` ...

mattsignorelli:
Could this be a bug in Bertz TPSA?

Give me an expression to write in FPP and I can evaluate it to 10th order or whatever

eiad-hamwi:
@mattsignorelli try this:
```
pc = 10^7
mc2 = 510998.95
z_offset = 1
z_offset * ((1+pz) * sqrt(pc^2 + mc2^2)/sqrt((pc*(1+pz))^2 + mc2^2) - 1)
```

This should be the exact value of `∆z` for a z_offset of 1m, using the value of `m_electron` in Bmad.

EDIT: I forgot a factor of `1+pz` in the numerator.

DavidSagan:
Etienne helped me diagnose the problem which was that Tao was not showing very small terms. Here "small" means that with reasonable values for the phase space coordinates the term would be small. This is fine for most purposes  but for this application not what is wanted. With the latest code there is now a `-noclean` switch for the `show taylor` command that will disallow the dropping of small terms.

DavidSagan:
And the problem with the constant term in z for patch elements turns out to be due to a difference between the electron mass used in Bmad (NIST 2022) and PTC (NIST 2014).

eiad-hamwi:
@mattsignorelli @DavidSagan, thanks for working with me on this. All is done and looks good now. 

Now the 10th-order patch map agrees to a precision of `5e-8` for each term.
Also implemented exact drift and solenoid with tests and 10th-order map agreement of `1e-8`.
Added cases for `ExactTracking` to reject other multipoles or combinations of patches, bends, multipoles.

src/modules/ExactTracking.jl
```diff
@@ -37,20 +37,103 @@ end
...
-                  * ((v[i,PXI]^2 + v[i,PYI]^2) - v[i,PZI] * (2 + v[i,PZI]) / gamsqr_0)
-                  / ( beta_0 * sqrt((1 + v[i,PZI])^2 + tilde_m^2) * work[i,1]
-                      * (beta_0 * sqrt((1 + v[i,PZI])^2 + tilde_m^2) + work[i,1])
-                    )
-                )
+    work[i,1] = 1 + v[i,PZI]                                 # 1+δp
+    work[i,2] = sqrt(work[i,1]^2 - v[i,PXI]^2 - v[i,PYI]^2)  # P_s
+    v[i,XI]   += v[i,PXI] * L / work[i,2]
+    v[i,YI]   += v[i,PYI] * L / work[i,2]
+    v[i,ZI]   -= work[i,1] * L * ( 1 / work[i,2] - sqrt( (p0c^2 + mc2^2) / ((p0c * work[i,1])^2 + mc2^2) ) )
```
mattsignorelli: @DavidSagan Is this numerically stable?

src/modules/ExactTracking.jl
```diff
@@ -37,20 +37,103 @@ end
...
+    # Temporaries
+    work[i,6] = work[i,4] * v[i,XI] / 2 + work[i,3] * (v[i,PXI] / ks + v[i,YI] / 2) + work[i,5] * v[i,PYI] / ks
+    work[i,7] = work[i,3] * (v[i,PYI] / 2 - ks * v[i,XI] / 4) + work[i,4] * v[i,PXI] / 2 - ks * work[i,5] * v[i,YI] / 4
+    work[i,8] = work[i,3] * (v[i,PYI] / ks - v[i,XI] / 2) + (work[i,4] * v[i,YI] - work[i,5] * v[i,PXI]) / ks
+    # Update
+    v[i,PYI]  = ks * work[i,5] * v[i,XI] / 4 - work[i,3] * (v[i,PXI] / 2 + ks * v[i,YI] / 4) + work[i,4] * v[i,PYI] / 2
+    v[i,XI]   = work[i,6]
+    v[i,PXI]  = work[i,7]
+    v[i,YI]   = work[i,8]
+    v[i,ZI]  += work[i,1] * L * ( sqrt(p0c^2 + mc2^2) / sqrt((p0c*work[i,1])^2 + mc2^2) - 1 / work[i,2] )
```
mattsignorelli: @DavidSagan  this too - is this numerically stable?

src/modules/ExactTracking.jl
```diff
@@ -37,20 +37,103 @@ end
...
+    if isnothing(winv)
+      @inbounds begin @FastGTPSA! begin
+      # No rotation case
+      v[i,XI] -= dx
+      v[i,YI] -= dy
+      
+      # Drift to face
+      v[i,XI] -= dz * v[i,PXI] / work[i,2]
+      v[i,YI] -= dz * v[i,PYI] / work[i,2]
+      v[i,ZI] += dz * work[i,1] / work[i,2] + dz*work[i,1]*sqrt((p0c^2+mc2^2)/((p0c*work[i,1])^2+mc2^2))
```
mattsignorelli: @DavidSagan

src/modules/ExactTracking.jl
```diff
@@ -37,20 +37,103 @@ end
...
+      v[i,PYI]  = winv[2,1]*work[i,5] + winv[2,2]*work[i,6] + winv[2,3]*work[i,2]
+      work[i,8] = winv[3,1]*work[i,5] + winv[3,2]*work[i,6] + winv[3,3]*work[i,2]
+      
+      # Drift length
+      work[i,9] = winv[3,1]*dx + winv[3,2]*dy + winv[3,3]*dz
+
+      # Drift to face
+      v[i,XI] -= work[i,7] * v[i,PXI] / work[i,8]
+      v[i,YI] -= work[i,7] * v[i,PYI] / work[i,8]
+      v[i,ZI] += work[i,7] * work[i,1] / work[i,8] + work[i,9]*work[i,1]*sqrt((p0c^2+mc2^2)/((p0c*work[i,1])^2+mc2^2))
```
mattsignorelli: @DavidSagan

src/modules/ExactTracking.jl
```diff
@@ -37,20 +37,103 @@ end
...
+
+# Rotation matrix
+function w_matrix(x_rot, y_rot, z_rot)
+  return ReferenceFrameRotations.angle_to_rot(-z_rot, x_rot, -y_rot, :ZXY)
+end
+
+# Inverse rotation matrix
+function w_inv_matrix(x_rot, y_rot, z_rot)
+  return ReferenceFrameRotations.angle_to_rot(y_rot, -x_rot, z_rot, :YXZ)
+end
```
mattsignorelli: Could you add some documentation on the choice of order of rotations and the reason for the negative signs you add?

test/bmad_maps/matrices.jl
mattsignorelli, line 1: We should not have matrices for any nonlinear trackings, only the maps, so we can remove this

test/ExactTracking.jl
```diff
@@ -0,0 +1,97 @@
...
+            dx = T(1)
+            dy = T(2)
+            dz = T(3)
+            winv = ExactTracking.w_inv_matrix(T(4),T(5),T(6))
+            return dx, dy, dz, winv
+        end
+
+        # Scalar patch jacobian
+        dx, dy, dz, winv = patch(Float64)
+        test_matrix(ExactTracking.patch!, M_patch_bmad, 10e6, 510998.95, dx, dy, dz, winv; tol=1e-7)
```
mattsignorelli: no need for any `test_matrix` in Exact tracking which nonlinear. In fact we should generalize your `coeffs_approx_equal` into a function similar to `test_matrix` called `test_map` which has the similar syntax but receives a  bmad `.jl` taylor map file as a string. Could you refactor your function into such a more general tool?

test/runtests.jl
```diff
@@ -19,6 +19,17 @@ function test_matrix(kernel, M_expected, args...; type_stable=VERSION >= v"1.11"
...
+    println("Kernel $(kernel) failed with norm $(norm(GTPSA.jacobian(v)[1:6,1:6] - scalar.(M_expected)))")
+    println("Expected: ")
+    for i in 1:6
+      println(scalar.(M_expected)[i,:])
+    end
+    println("Received: ")
+    for i in 1:6
+      println(GTPSA.jacobian(v)[i,:])
+    end
+  end
```
mattsignorelli: This is not necessary - the test failure will print such a statement
```suggestion
```

mattsignorelli (changes requested):
Need @DavidSagan  to check the modified exact drift, and your exact solenoid, for numerical stability. Then just some minor changes

src/modules/ExactTracking.jl
mattsignorelli: Can we instead call `exact_drift!` here so we do not have repetitive code?

src/modules/ExactTracking.jl
mattsignorelli: Can we instead just call `exact_drift!` here so we do not have repetitive code?

test/runtests.jl
eiad-hamwi: Yep, this was just a diagnostic I forgot to remove.

src/modules/ExactTracking.jl
eiad-hamwi: We can use `exact_drift!` here.

src/modules/ExactTracking.jl
eiad-hamwi: Here we can't use `exact_drift!` because the drift length of an off-axis particle is different from that of the reference particle due to the reference frame rotation, so the `z` coordinate has a different expression.

mattsignorelli: What am I missing exactly? The code here looks identical to `exact_drift!` just with `L=-dz`?

DavidSagan: With a patch the User can set `L` so `L` is not necessarily related to `dz`.

eiad-hamwi: In the patch with rotation, the two terms in the last line have different length coefficients: `work[i,7]` and `-work[i,9]`

src/modules/ExactTracking.jl
eiad-hamwi: Done

test/ExactTracking.jl
eiad-hamwi: We can open a new issue for that.

src/modules/ExactTracking.jl
eiad-hamwi: Small `pz` makes `v[i,ZI]` changes unstable.

Reverted to previous drift code which by Dan which avoids this issue.

test/ExactTracking.jl
eiad-hamwi: Fixed

src/modules/ExactTracking.jl
eiad-hamwi: Fixed numerical instability `v[i,ZI]` for small `pz`. Rest of coordinates are stable for small phase space coords.

eiad-hamwi:
All issues addressed. 
Note that I added a diagnostic print statement which outputs the first incorrect GTPSA coefficient and its error since the `@test` macro will miss this.

mattsignorelli:
This looks great! We will need a review from @DavidSagan to check carefully for numerical stability. I also see some places where we could gain some minor performance optimizations (e.g. going from charge and Brho_ref to gammasqr0 and stuff).  

The drift in the patch (p0c and mass) ideally should receive the same parameters as the drifting part in the bend and solenoid and drift (gamsqr_0, tilde_m, beta). @DavidSagan @dtabell we need to discuss a regular convention for the drift parameters. Right now it seems to be gamsqr_0, tilde_m, and beta_0, but I'm not sure those are the most convenient anymore with the bunch struct storing Brho_ref.  The patch should then use these same parameters.

Another point - I will be adding in a macro to allow for your own temporary variable names (instead of work[i,j]) for readability. In the meantime, when you get the chance, could you add in some appropriate variable names as comments next to each work[i,j] which i can then change later?

eiad-hamwi:
Purely in terms of minimal number of parameters, we only need one parameter such as `tilde_m == mc2 / p0c` since the following relations hold:
```
gamsqr_0 == 1 + 1/tilde_m^2
beta_0 == 1/sqrt(1 + tilde_m^2)
```

DavidSagan (approved):
No equations with round off problems spotted.

src/modules/ExactTracking.jl
```diff
@@ -38,19 +39,128 @@ end
...
+The inverse matrix reverses the order of operations and their signs.
+
+
+Arguments:
+- `x_rot::Real`: Rotation angle around the x-axis.
+- `y_rot::Real`: Rotation angle around the y-axis.
+- `z_rot::Real`: Rotation angle around the z-axis.
+
+Returns:
+- `Matrix{Float64}`: Rotation matrix.
```
mattsignorelli:
```suggestion
- `SMatrix{Float64}`: Rotation matrix.
```

src/modules/ExactTracking.jl
```diff
@@ -38,19 +39,128 @@ end
...
+Furthermore, in ReferenceFrameRotations, the rotation angles around Y and Z axes 
+are defined as negative of the SciBmad `y_rot` and `z_rot`.
+
+The inverse matrix reverses the order of operations and their signs.
+
+
+Arguments:
+- `x_rot::Real`: Rotation angle around the x-axis.
+- `y_rot::Real`: Rotation angle around the y-axis.
+- `z_rot::Real`: Rotation angle around the z-axis.
```
mattsignorelli:
```suggestion
- `x_rot::Number`: Rotation angle around the x-axis.
- `y_rot::Number`: Rotation angle around the y-axis.
- `z_rot::Number`: Rotation angle around the z-axis.
```

according to the [docs](https://juliaspace.github.io/ReferenceFrameRotations.jl/stable/lib/library/#ReferenceFrameRotations.angle_to_rot-Tuple{Number,%20Symbol}) these do not need to be real but `Number`. The `TPS` type for example only subtypes `Real` and not `Number`

test/bmad_maps/comp.jl
```diff
@@ -0,0 +1,112 @@
...
+    else
+        if switch
+            return false
+        else
+            error("Invalid monomial index: $idx, $current_idx")
+        end
+    end
+end
+
+using BeamTracking
```
mattsignorelli: Should this line inclusive and below be in a separate file?

test/bmad_maps/comp.jl
```diff
@@ -0,0 +1,112 @@
+using GTPSA
```
mattsignorelli: I found the GTPSA function we want: 
https://bmad-sim.github.io/GTPSA.jl/stable/devel/#GTPSA.mad_tpsa_dif!

The destination TPS contains the errors and then we can compute the error from that from the `normTPS` for example or go order by order

test/runtests.jl
```diff
@@ -29,5 +29,65 @@ function test_matrix(kernel, M_expected, args...; type_stable=VERSION >= v"1.11"
...
+      @test_opt BeamTracking.launch!(kernel, v, work, args...)
+  end
+  # 3) No allocations
+  if no_allocs
+      @test @ballocated(BeamTracking.launch!($kernel, $v, $work, $args...)) == 0
+  end
+end
+
+#  Coefficient-wise approximate equality
+function coeffs_approx_equal(v_expected, v_calculated, ϵ)
```
mattsignorelli, line 85: Oh is `comp.jl` not needed?

ext/BeamTrackingBeamlinesExt/exact.jl
```diff
@@ -0,0 +1,76 @@
...
+    elseif isactive(bmultipoleparams) # Straight multipole
+      if haskey(bmultipoleparams.bdict, 0) # Solenoid
+        if L < 1e-6
+          error("Exact thin-lens solenoid not yet implemented (L = 0)")
+        else
+          Ks = get_thick_strength(bmultipole.bdict[0], L, bunch.Brho_ref)*chargeof(bunch.species)
+          p0c = bunch.Brho_ref*C_LIGHT*chargeof(bunch.species)
+          tilde_m = massof(bunch.species)/p0c
+          gamsqr_0 = 1 + 1/tilde_m^2
+          beta_0 = sqrt(1 - 1/gamsqr_0)
```
mattsignorelli: Note that:
```
tilde_m = mc/(gamma*m*v) = 1/(gamma*beta)
gamsqr_0 = 1+gamma^2*beta^2
```

So I suggest:
```suggestion
          beta_gamma_0 = bunch.Brho_ref*chargeof(bunch.species)*C_LIGHT/massof(bunch.species)
          tilde_m = 1/beta_gamma_0
          gamsqr_0 = 1+beta_gamma_0^2
          beta_0 = beta_gamma_0/sqrt(gamsqr_0)
```

We should move this functionality  into `src/utils.jl` for generality and to apply `@FastGTPSA` to. Function should return `beta_0`, `tilde_m`, and `gamsqr_0`

mattsignorelli: done

test/runtests.jl
mattsignorelli: deleted comp.jl

test/bmad_maps/comp.jl
mattsignorelli: your coefficient comparison is good

test/bmad_maps/comp.jl
mattsignorelli: deleted

src/modules/ExactTracking.jl
```diff
@@ -38,19 +39,138 @@ end
...
+
+# Rotation matrix
+"""
+  w_matrix(x_rot, y_rot, z_rot)
+
+Constructs a rotation matrix based on the given Bryan-Tait angles.
+
+Bmad/SciBmad follows the MAD convention of applying z, x, y rotations in that order.
+Furthermore, in ReferenceFrameRotations, the rotation angles around Y and Z axes 
+are defined as negative of the SciBmad `y_rot` and `z_rot`.
```
mattsignorelli: Is this true? Looking at the documentation for ReferenceFrameRotations and I thought SciBmad is the same?

test/runtests.jl
```diff
@@ -29,5 +29,65 @@ function test_matrix(kernel, M_expected, args...; type_stable=VERSION >= v"1.11"
...
+  mod = Module()
+  Base.include(mod, bmad_map_file)
+  D = getfield(mod, :d_z)
+  v_expected = getfield(mod, :v_z)
+
+  # Construct test input variables
+  n_temps = BeamTracking.MAX_TEMPS(parentmodule(kernel).TRACKING_METHOD())
+  v = transpose(@vars(D))
+  work = zeros(eltype(v), 1, n_temps)
+  args = to_type.(args, TPS64{D})
```
mattsignorelli: What is the point of this line and `to_type`?

mattsignorelli:
I cleaned up the code, just some minor left-over points above. Also we do need some regression test for `BeamlinesExt` , e.g. using the HSR

ext/BeamTrackingBeamlinesExt/exact.jl
```diff
@@ -0,0 +1,70 @@
...
+    if isactive(bmultipoleparams) # Combined function bend
+      error("Exact bend cannot be used with multipole elements")
+    else
+    end
+  elseif isactive(bmultipoleparams) # Straight multipole
+    if haskey(bmultipoleparams.bdict, 0) # Solenoid
+      if L == 0
+        error("Exact thin-lens solenoid not yet implemented (L = 0)")
+      else
+        Ks = get_thick_strength(bmultipole.bdict[0], L, bunch.Brho_ref)*chargeof(bunch.species)
```
mattsignorelli: Why thick strength * charge? doesn't Brho_ref capture the sign or am I confused

mattsignorelli:
Also per the code coverage report we should have a patch test without rotation since there is a different branch used for that inside the kernel

I'd like to merge this soon so that I can merge my updates (spin, new KernelChain thing, local variable names, etc) after

src/modules/ExactTracking.jl
eiad-hamwi: Yes, that's why the map comparison agrees. Look at `angle_to_dcm(θ::T, rot_seq::Symbol)` in https://github.com/JuliaSpace/ReferenceFrameRotations.jl/blob/master/src/conversions/angle_to_dcm.jl

ext/BeamTrackingBeamlinesExt/exact.jl
eiad-hamwi: Nope, you are correct, it should just be thick strength. This remained from when I was working with `p0c` and not `Brho_ref`.

src/modules/ExactTracking.jl
DavidSagan: @eiad-hamwi Looking at your reference it looks like all three rotations ReferenceFrameRotations are negative of SciBmad.

eiad-hamwi: @DavidSagan That would be true for the definition of the rotation matrix itself, but the transformation for the W matrix in Bmad uses a convention that rotates around the -X axis, rather than the X axis itself, c.f. equations 16.2 and 16.3 in the Bmad manual. So it depends on what you want to call a rotation around the X axis

DavidSagan: No. SciBmad `x_rot` is negative Bmad `y_pitch`.

test/runtests.jl
eiad-hamwi: I want to pass the kernel arguments as floats and then convert to TPS because the descriptors are defined locally in the Bmad map files.

`to_type` is for broadcasting, but I guess I could just use the `.` syntax

src/modules/ExactTracking.jl
DavidSagan: In other words, in SciBmad, `x_rot`, `y_rot`, and `z_rot` are all positive rotations around the `x`, `y`, and `z` axes.

test/runtests.jl
mattsignorelli: No I mean, there are two cases - when the parameters are floats and when the parameters are TPSAs. In the linear tracking tests I test both, and `test_matrix` just uses whatever arguments its given. But here you force all parameters to be `TPSAs`? Should remove this line and have tests for both

eiad-hamwi: I don't see the point of testing both: if it works on TPSA parameters, it works on floats. What am I missing?

mattsignorelli: While this is true for correctness, it is not true for allocations. For example if someone else uses `test_map` and sees allocations, it won't be immediately clear if that's the result of a mishandled TPSA operation (which always allocates) or if something else went wrong with regular numbers (e.g. maybe they allocated a temporary array).

You are correct that passing the tests with just TPSA is good, but since your `test_map` will be used by many people and will likely be a useful diagnostic tool in the above cases, in theory there should be tests for both

eiad-hamwi: Gotcha, added this.

eiad-hamwi:
@mattsignorelli as the commit says, here are the corrections / changes:

1. Corrected typo in `patch!` Z update in no rotations branch (self-explanatory)
2. Explicitly inlined `exact_drift!` into `patch!` to avoid allocations.

   -  `exact_drift!` was making 80 allocations when `patch!` called it in the no rotations branch.
   -  Using the `@inline` macro in front of `exact_drift!` did not fix this issue with allocations.
   -  This is likely due to some incompatibility between `@inline` and `@makekernel`

3. Added `patch_norot` test to cover other patch branch (self-explanatory)
4. Added tests for scalar parameters (self-explanatory)

mattsignorelli:
>     2. Explicitly inlined `exact_drift!` into `patch!` to avoid allocations.
>        
>        * `exact_drift!` was making 80 allocations when `patch!` called it in the no rotations branch.
>        * Using the `@inline` macro in front of `exact_drift!` did not fix this issue with allocations.
>        * This is likely due to some incompatibility between `@inline` and `@makekernel`


Allocations cannot be related to the inlining here because everything is type stable. Also `@makekernel` automatically applies `@inline`. I suspect it could be related to either `@FastGTPSA` or setting `v = exact_drift!`...

mattsignorelli:
Also, it seems `dz`  is not used when there's no rotations?

DavidSagan:
Note: In AcceleratorLattice patch rotation names are `x_rot`, etc. and not `dx_rot`. Personally, I do like `x_rot` better.

eiad-hamwi:
@mattsignorelli actually, you just reminded me that `L` is set by the user, which means that `L` and `dz` can be different, so that we cannot actually use `exact_drift!`, just like in the branch with rotations.

src/modules/ExactTracking.jl
eiad-hamwi: @DavidSagan corrected.

eiad-hamwi:
> Note: In AcceleratorLattice patch rotation names are `x_rot`, etc. and not `dx_rot`. Personally, I do like `x_rot` better.

@DavidSagan I think @mattsignorelli and I agree with you. Why don't you open an issue in Beamlines for it?

mattsignorelli:
> > Note: In AcceleratorLattice patch rotation names are `x_rot`, etc. and not `dx_rot`. Personally, I do like `x_rot` better.
> 
> @DavidSagan I think @mattsignorelli and I agree with you. Why don't you open an issue in Beamlines for it?

See the discussion here: https://github.com/bmad-sim/Beamlines.jl/pull/28

DavidSagan:
Issue submitted.

mattsignorelli:
> > Note: In AcceleratorLattice patch rotation names are `x_rot`, etc. and not `dx_rot`. Personally, I do like `x_rot` better.
> 
> @DavidSagan I think @mattsignorelli and I agree with you. Why don't you open an issue in Beamlines for it?

I do not agree

https://github.com/bmad-sim/Beamlines.jl/pull/28#issuecomment-2849340563

mattsignorelli:
> > Note: In AcceleratorLattice patch rotation names are `x_rot`, etc. and not `dx_rot`. Personally, I do like `x_rot` better.
> 
> @DavidSagan I think @mattsignorelli and I agree with you. Why don't you open an issue in Beamlines for it?

eiad-hamwi:
@mattsignorelli are you not going to merge this branch?

mattsignorelli:
Oops I did not mean to close

mattsignorelli:
Seems like all that's left are 
1. Fixing the sign convention for patch to agree with SciBmad per David's comments
2. add regression test for HSR
3. We should have a regression test for the exact solenoid in the BeamlinesExt too, if the HSR does not include any solenoid

eiad-hamwi:
@mattsignorelli 

1. Already fixed in last commit
2. HSR uses some combined function magnets, so we cannot do that yet
3. What kind of regression tests are you thinking of? 

We’re already sure of the map to 10th order, so do you mean just comparing the output of tracking a particle/bunch through an open lattice of these elements?

src/modules/ExactTracking.jl
```diff
@@ -38,19 +39,140 @@ end
...
+
+# Rotation matrix
+"""
+  w_matrix(x_rot, y_rot, z_rot)
+
+Constructs a rotation matrix based on the given Bryan-Tait angles.
+
+Bmad/SciBmad follows the MAD convention of applying z, x, y rotations in that order.
+Furthermore, in ReferenceFrameRotations, the rotation angles are defined as negative 
+of the SciBmad rotation angles `z_rot`, `y_rot`, and `z_rot`.
```
mattsignorelli:
```suggestion
of the SciBmad rotation angles `x_rot`, `y_rot`, and `z_rot`.
```

mattsignorelli:
OK great!

I just mean tests for the exact.jl file in BeamlinesExt. We need coverage for your added lines there. You can use the same parameters and maps as you did for the lower level kernel tests. This just makes sure that given an element in Beamlines, the correct lower level kernel and values are called

eiad-hamwi:
Done

mattsignorelli (approved):

mattsignorelli:
The error is because I need to release the new `Beamlines`. I will do so now and then if all tests pass this will be merged

eiad-hamwi:
Alriiight :D

mattsignorelli:
@eiad-hamwi  can you bump the Beamlines compat version to the latest ( v0.2.2 just released) on this branch  and see if the tests here pass

eiad-hamwi:
Done now

mattsignorelli:
Thanks for all of your work!

eiad-hamwi:
My pleasure!

PR #81, opened 2025-05-14
Kernelab

mattsignorelli:


mattsignorelli:
This PR implements KernelAbstractions into the package, as well as my generic parallel kernel launching function `launch!`. With no `groupsize` specified when using KernelAbstractions, CPU implementation gives similar performance as my original implementation. 

Also included are keyword arguments whose access have been elevated to the `track!`. These allow people to customize the `groupsize`, `multithread_threshold`, `use_KA`, and `use_explicit_SIMD`. When using a keyword argument, the dispatch on the tracking method does require one allocation, however NO performance loss time-wise is observed. Such an allocation was expected before anyways.

Testing on the GPU is ready

mattsignorelli:
Needed is some `@makekernel` macro to create both the `@kernel` and its inside function

Issue #93, opened 2025-06-05
Next release to-do


mattsignorelli:
- [ ] New Beamlines release
- [ ] update compat here
- [ ] Merge #79 
- [ ] merge branch spin (including spin, state, `KernelChain`)
- [ ] remove work array, make `@makekernel` only do `@FastGTPSA` not `@FastGTPSA!`, and no allocs tests for GTPSA map

One issue I just realized is that doing something like 

`old_x = v[i,XI]`

does not allocate a new TPS for the temporary by default, unless the code is wrapped in `@FastGTPSA`

PR #94, opened 2025-06-06
Adding extensive documentation and readability improvements to the core module

ndwang:
I added documentation for the core module, especially doc strings for all the functions.

Two changes to the code:

constants for coordinate indexing now lives in types.jl instead of kernel.jl, right next to struct Bunch definition.
Refactored launch! to avoid deep nested if-else branching. Functionally identical, just much easier to read and follow.


mattsignorelli:
Thanks. I will look at this after #93 . I made a lot of improvements and changes to `launch!` in my current branch that might address some of these too

ndwang:
This is a repost of a previous PR. I moved it to a separate branch to make way for another PR.

PR #95, opened 2025-06-06
Implemented FieldTracking and explicit RK4

ndwang:
Implemented features

1. src/modules/FieldTracking.jl
Implemented kernel for FieldTracking for an arbitrary field function using DifferentialEquations.jl. Currently only supports static electric field.

2. src/modules/RungeKuttaTracking.jl
Implemented explicit 4th order Runge-Kutta tracking.

3. test/FieldTracking,jl
Three test cases for FieldTracking: Equation of motion, single particle tracking, and multi-particle tracking.

4. test/KernelEvaluation.jl
A general benchmarking tool to measure the performance of any tracking kernel.


Project.toml
```diff
@@ -4,10 +4,13 @@ authors = ["mattsignorelli <mgs255@cornell.edu> and contributors"]
 version = "0.1.0"
 
 [deps]
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
```
mattsignorelli:
```suggestion
```
`BenchmarkTools` should just be an `extra` for files in `test` to depend on, not a dependency of `BeamTracking` itself

Project.toml
```diff
@@ -19,10 +22,13 @@ BeamTrackingBeamlinesExt = "Beamlines"
 
 [compat]
 Beamlines = "0.2.1"
+BenchmarkTools = "1.6.0"
```
mattsignorelli:
```suggestion
```
Likewise from above comment

Project.toml
```diff
@@ -4,10 +4,13 @@ authors = ["mattsignorelli <mgs255@cornell.edu> and contributors"]
 version = "0.1.0"
 
 [deps]
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 GTPSA = "b27dd330-f138-47c5-815b-40db9dd9b6e8"
 KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
```
mattsignorelli: We should only need `SciMLBase`, right? ODE is a very heavy dependency so they recommend only depending on `SciMLBase`

src/BeamTracking.jl
```diff
@@ -1,3 +1,9 @@
+"""
+    BeamTracking
+
+A high-performance particle beam tracking package for accelerator physics simulations.
+Provides both linear and exact tracking methods with optimized memory layouts and parallel processing.
```
mattsignorelli: Many more tracking methods than just exact and linear

src/modules/FieldTracking.jl
```diff
@@ -0,0 +1,69 @@
...
+- `solver`: ODE solver to use
+- `solver_params`: Additional parameters for the solver
+"""
+@makekernel function field_track!(i, v, work, L, field_func, field_params, solver, solver_params)
+    @inbounds begin
+        # Initial state vector
+        u0 = view(v, i, :)
+        
+        # Set up and solve the ODE
+        prob = ODEProblem(field_system!, u0, (0.0, L), (field_func, field_params))
```
mattsignorelli: Shouldn't the ODEProblem be set up outside of the kernel so that the kernel itself just calls `solve`?

test/KernelEvaluation.jl
```diff
@@ -0,0 +1,82 @@
+using BeamTracking
+using BeamTracking: get_N_particle, runkernel!, MAX_TEMPS, soaview
+using BenchmarkTools
+using SciMLBase, OrdinaryDiffEq
+
+"""
+    evaluate_kernel_performance(bunch, kernel, args...; n_runs=10)
```
mattsignorelli, line 8: This is very useful!

mattsignorelli:
This is some good stuff, just a couple of points

src/modules/RungeKuttaTracking.jl
```diff
@@ -0,0 +1,90 @@
...
+- `field_func`: Function that returns the field. Must be of the form `field_func(u, t, params)`.
+    Return value should be [px, Ex, py, Ey, pz, Ez].
+- `params`: Additional parameters for the field function
+"""
+function rk4_step!(i, u, work, t, h, field_func, params)
+    # Get views into work matrix for RK4 stages
+    k1 = view(work, i, 1:6)    # First 6 elements for stage 1
+    k2 = view(work, i, 7:12)   # Next 6 elements for stage 2
+    k3 = view(work, i, 13:18)  # Next 6 elements for stage 3
+    k4 = view(work, i, 19:24)  # Last 6 elements for stage 4
```
mattsignorelli: Pinging @obeznosov-LANL @jackelley

src/modules/FieldTracking.jl
mattsignorelli: As in anything from the `DifferentialEquations` suite I imagine might override `runkernel` and use its `EnsembleProblem`? And I think we should have default `solver_params` - especially the ones that only save the final state

mattsignorelli: But using your `evaluate_kernel_performance` we could check both :)

src/BeamTracking.jl
ndwang: They don't exist yet. This should be updated as new things are implemented

src/modules/FieldTracking.jl
ndwang: Running an `ODEProblem` for each particle is a good baseline. We can benchmark this against `EnsembleProblem` and others later
`EnsembleProblem` needs a different setup around it. It'll mostly be different codes

ndwang: > Shouldn't the ODEProblem be set up outside of the kernel so that the kernel itself just calls `solve`?

To set up `ODEProblem` you need `u0`, meaning this is already on a single particle. It's possible to have another layer on top that sets up `ODEProblem` before `field_track!`, but it's functionally the same.

An alternative is to set up all `ODEProblem`s for the whole bunch, store them in a big array, and pass it to the kernel. But I suspect this is a waste of memory because they are only used once yet live until all particles are tracked. GPUs need separate considerations of course.

ndwang: >  we should have default `solver_params` - especially the ones that only save the final state

I am experimenting with this. The default here disables all saves.
https://github.com/ndwang/BeamTracking.jl/blob/main/test/KernelEvaluation.jl#L59

However, `@makekernel` doesn't support default arguments. So any default setting will have to be outside the kernel.

ndwang:
RK4 now has 0 allocations
```
julia> evaluate_rk4_track_performance()
Dict{String, Real} with 4 entries:
  "min_allocs" => 0
  "success"    => true
  "min_time"   => 889100.0
  "min_memory" => 0
```

ndwang:
Compare to the previous implementation
```
julia> evaluate_rk4_track_performance()
Dict{String, Real} with 4 entries:
  "min_allocs" => 60000
  "success"    => true
  "min_time"   => 1.0898e6
  "min_memory" => 3360000
```

Forcing OrdinaryDiffEq's RK4 to use the same fixed step size and no saves
```
evaluate_field_track_performance(solver=RK4(),solver_params=(save_everystep=false,save_start=false,save_end=true,dense=false,calck=false,adaptive=false,dt=0.1))
Dict{String, Real} with 4 entries:
  "min_allocs" => 135000
  "success"    => true
  "min_time"   => 1.4491e7
  "min_memory" => 7456000
```

ndwang:
Kernel interfaces updated to the latest standard

src/modules/FieldTracking.jl
```diff
@@ -0,0 +1,65 @@
...
+- `i`: Particle index
+- `b`: BunchView containing particle coordinates
+- `L`: Drift length
+- `field_func`: Function that returns the field at a given position (x, y, z)
+- `field_params`: Additional parameters for the field function
+- `solver`: ODE solver to use
+- `solver_params`: Additional parameters for the solver
+"""
+@makekernel function field_track!(i, b::BunchView, L, field_func, field_params, solver, solver_params)
+    @inbounds begin
```
mattsignorelli: `@makekernel` automatically wraps the code with an `inbounds`, so this is not necessary

src/modules/FieldTracking.jl
mattsignorelli: I suspect `EnsembleProblem` with overridden `runkernels!` will be faster here

src/modules/FieldTracking.jl
```diff
@@ -0,0 +1,65 @@
...
+        # Initial state vector
+        u0 = view(b.v, i, :)
+        
+        # Set up and solve the ODE
+        prob = ODEProblem(field_system!, u0, (0.0, L), (field_func, field_params))
+        sol = solve(prob, solver; reltol=1e-8, abstol=1e-8, solver_params...)
```
mattsignorelli: Why these tolerances specifically?

src/modules/RungeKuttaTracking.jl
```diff
@@ -0,0 +1,89 @@
...
+    RungeKuttaFieldTracking
+
+Module implementing particle tracking through arbitrary electromagnetic fields using a 4th order Runge-Kutta method.
+"""
+
+# Define the RungeKutta tracking method
+struct RungeKutta end
+
+# Number of temporaries needed for a single particle
+MAX_TEMPS(::RungeKutta) = 24  # Number of RK4 stages
```
mattsignorelli:
```suggestion
```

No more work array

src/modules/RungeKuttaTracking.jl
```diff
@@ -0,0 +1,89 @@
...
+- `i`: Particle index
+- `u`: State vector [x, px, y, py, z, pz]
+- `work`: Work matrix (n_particles × 24)
+- `t`: Current time
+- `h`: Step size
+- `field_func`: Function that returns the field. Must be of the form `field_func(u, t, params)`.
+    Return value should be [px, Ex, py, Ey, pz, Ez].
+- `params`: Additional parameters for the field function
+"""
+function rk4_step!(i, u, work, t, h, field_func, params)
```
mattsignorelli: This function + docs is not compatible with new framework

test/FieldTracking.jl
```diff
@@ -0,0 +1,55 @@
...
+        @test isapprox(bunch.v[1,2], 1.0, rtol=1e-5)  # px = t
+    end
+    
+    # Test field_track! with multiple particles
+    @testset "Multiple Particle Tracking" begin
+        # Create multiple particles
+        bunch = Bunch(zeros(3,6))
+        bunch.v[2,1] = 1.0
+        bunch.v[3,2] = 1.0
+        work = zeros(eltype(bunch.v), get_N_particle(bunch), MAX_TEMPS(ele.tracking_method))
```
mattsignorelli: No more work

test/KernelEvaluation.jl
```diff
@@ -0,0 +1,82 @@
...
+- `success`: Boolean whether the tracking was successful
+
+"""
+function evaluate_kernel_performance(bunch, kernel, args...; n_runs=10)
+    n_particles = get_N_particle(bunch)
+
+    # Get the tracking method from the kernel's module
+    tracking_method = parentmodule(kernel).TRACKING_METHOD()
+    n_temps = MAX_TEMPS(tracking_method)
+    work = zeros(eltype(bunch.v), n_particles, n_temps)
```
mattsignorelli: no more work

mattsignorelli (changes requested):

src/modules/RungeKuttaTracking.jl
ndwang: This is an internal function, not a kernel

ndwang:
@mattsignorelli  kernels no longer require work != kernels are forbidden from using work

ndwang:
Seems like `KernelChain` changes the code behavior in a significant way

no allocation implementation using work array
```
julia> evaluate_rk4_track_performance()
Dict{String, Real} with 4 entries:
  "min_allocs" => 11
  "success"    => true
  "min_time"   => 904300.0
  "min_memory" => 304
```

naive implementation using local variables
```
julia> evaluate_rk4_track_performance()
Dict{String, Real} with 4 entries:
  "min_allocs" => 60011
  "success"    => true
  "min_time"   => 482100.0
  "min_memory" => 3360288
```

src/modules/FieldTracking.jl
ndwang: Removed default tolerance. Moving this to the wrapper

ndwang:
All the interfaces are updated. `EnsembleProblem` will be implemented separately and benchmarked

src/modules/FieldTracking.jl
```diff
@@ -0,0 +1,60 @@
...
+
+"""
+    FieldTracking
+
+Module implementing particle tracking through arbitrary electromagnetic fields using DifferentialEquations.jl.
+"""
+module FieldTracking
+using ..BeamTracking
+using ..BeamTracking: @makekernel, BunchView
+using SciMLBase
```
mattsignorelli, line 11:
```suggestion
using ..SciMLBase
```

src/modules/FieldTracking.jl
```diff
@@ -0,0 +1,60 @@
...
+        # Initial state vector
+        u0 = view(b.v, i, :)
+        
+        # Set up and solve the ODE
+        prob = ODEProblem(field_system!, u0, (0.0, L), (field_func, field_params))
+        sol = solve(prob, solver; solver_params...)
+        
+        # Update final coordinates by assigning each component
+        u0 .= sol.u[end]
+    end
```
mattsignorelli:
```suggestion
```

src/modules/FieldTracking.jl
```diff
@@ -0,0 +1,60 @@
...
+- `i`: Particle index
+- `b`: BunchView containing particle coordinates
+- `L`: Drift length
+- `field_func`: Function that returns the field at a given position (x, y, z)
+- `field_params`: Additional parameters for the field function
+- `solver`: ODE solver to use
+- `solver_params`: Additional parameters for the solver
+"""
+@makekernel function field_track!(i, b::BunchView, L, field_func, field_params, solver, solver_params)
+    begin
```
mattsignorelli:
```suggestion
```

src/modules/RungeKuttaTracking.jl
```diff
@@ -0,0 +1,69 @@
...
+- `i`: Particle index
+- `b`: BunchView containing particle coordinates
+- `t_span`: Time span [t_start, t_end]
+- `field_func`: Function that returns the field. Must be of the form `field_func(u, t, params)`.
+    Return value should be [px, Ex, py, Ey, pz, Ez].
+- `params`: Additional parameters for the field function
+- `n_steps`: Number of integration steps
+"""
+@makekernel function rk4_track!(i, b::BunchView, t_span, field_func, params, n_steps)
+    begin
```
mattsignorelli:
```suggestion
```

src/modules/RungeKuttaTracking.jl
```diff
@@ -0,0 +1,69 @@
...
+        
+        # Integration step size
+        h = (t_span[2] - t_span[1]) / n_steps
+
+        t = t_span[1]
+        # Perform integration steps
+        for _ in 1:n_steps
+            rk4_step!(u, t, h, field_func, params)
+            t += h
+        end
```
mattsignorelli:
```suggestion
```

mattsignorelli:
> All the interfaces are updated. EnsembleProblem will be implemented separately and benchmarked

OK. In this case we can merge this now and wait for this benchmarking. Do you plan to add a `FieldMapParams` to `Beamlines`?

Other than that just a couple of minor things:

- [ ] minor edits above
- [x] I noticed you use four spaces for your tabs, but the rest of the SciBmad code base uses 2 spaces. Could you modify the files you added to follow this convention?
- [x] Per the code coverage report can you add tests for `RungeKuttaTracking`?

PR #102, opened 2025-06-13
Spinchain

mattsignorelli:


mattsignorelli:
This PR implements the following:
- Quaternion array `q` added to `Bunch`
- State array `state` added to `Bunch`
- New `BunchView` struct containing `v`, `q`, and `state` which is now the only required argument for all kernels other than `i`, e.g. all kernels now are of format `foo(i, b::BunchView, any_args...)`
- `BunchView` is GPU compatible via `Adapt.jl`
- Work array has been removed
- All kernels when using KernelAbstractions go through a single `generic_kernel!`
- `@makekernel` simplified bc ^, now only applies `@inbounds` (default true), `@FastGTPSA` (default false) to  the entire function given kwargs `inbounds` and `fastgtpsa` e.g. `@makekernel fastgtpsa=true foo(i, b::BunchView, args...)...`. `@makekernel` also makes sure that there is no return statement and always adds a `return nothing` to the end following KernelAbstractions convention
- New `KernelCall` structure to specify a kernel call , e.g. with `foo` above would be `KernelCall(foo, (any_args...,))`
- New `KernelChain` construct which is just a tuple of `KernelCalls`, and appropriate `push` function so that kernel chains can be constructed on a per-element basis, and run at the end of the element. The compiler appears able to elude the type instability in this specific case. In the future will likely make separate function to construct kernel chain only for an element in case that is desired e.g. for CUDA graph stuff (though `BitsBeamline` should still be the way to go due to memory constaints).
- `launch!` only accepts `KernelChain`, in the way: `launch!(b::BunchView, kc::KernelChain; kwargs...)`
- With the new `BunchView`, the compiler seems reluctant to autovectorize. Therefore on the CPU the default is now `use_KA=false` and `use_explicit_SIMD=true`. For GPU `use_KA=true` always
- `runkernels!` replaces `runkernel`, now accepting a `BunchView` and `KernelChain`, as well as the `i` argument specifying if inside of a launch already or not (nothing)
- Tests modified to play nicely with new package structure

mattsignorelli:
Due to a bug in SIMD.jl https://github.com/eschnett/SIMD.jl/issues/140 , `use_explicit_SIMD` defaults to `false` now, and `use_KA=true` is always the default on both CPU and GPU

PR #109, opened 2025-06-17
Purely positive `p0c`, `brho`

eiad-hamwi:
The convention for `Brho` and `p0c` is that they are both positive for +/- charge, so I've made this change in internal calculations.

In the future, we may want to store the direction (up/down) of the holding dipole field in a different place.

### DIFF
--- a/src/modules/ExactTracking.jl
+++ b/src/modules/ExactTracking.jl
@@ -447,7 +447,7 @@ function w_inv_matrix(x_rot, y_rot, z_rot)
 end
 
 function drift_params(species::Species, Brho)
-  beta_gamma_0 = BeamTracking.calc_beta_gammma(species, Brho)
+  beta_gamma_0 = BeamTracking.calc_beta_gamma(species, Brho)
   tilde_m = 1/beta_gamma_0
   gamsqr_0 = @FastGTPSA 1+beta_gamma_0^2
   beta_0 = @FastGTPSA beta_gamma_0/sqrt(gamsqr_0)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -130,12 +130,12 @@ massof(s::Species) = s.mass
 chargeof(s::Species) = s.charge
 
 # Particle energy conversions =============================================================
-calc_Brho(species::Species, E) = @FastGTPSA sqrt(E^2-massof(species)^2)/C_LIGHT/chargeof(species)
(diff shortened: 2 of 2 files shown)


mattsignorelli:
`p0c` and `beta_gamma` should definitely always be positive, but the couldn't `Brho_ref` be negative? @DavidSagan ?

mattsignorelli:
For now let's just set `Brho_ref` (maybe rename later to `rigidity` or something) to a negative value for negatively-charged particles. Then everything is still fully consistent. If you use an always-positive definition, then by changing the charge of the particle going through, you flip the fields of the entire accelerator - things become inconsistent

Maybe `Beamlines` will later allow people to optionally set a reference species if they would like to shield the user from this, but I think it is important that internally the code remains consistent.

eiad-hamwi:
Personally, I agree with that interpretation of `Brho_ref`, especially because it directly tells us the orientation of the holding fields.

I think it was Georg who had a stronger opinion to avoid confusing people because of its conventional interpretation as a purely positive quantity.

eiad-hamwi:
@mattsignorelli Georg on Friday suggested that maybe we can and should switch from using `Brho` to `poq` "p over q" which does not care about any conventions and can be positive or negative by definition. @DavidSagan do you prefer this since you mentioned you think that `Brho` is antiquated? What do you guys think?

mattsignorelli:
I think I like `rigidity_ref` or `ref_rigidity`. Then we can provide some fluff at the higher level to handle the sign of this value stored in the `Bunch`/`Beamline`.

eiad-hamwi:
I think `rigidity` and `Brho` are usually considered synonyms of each other, so that will still break with the purely positive convention. 

Unless you're just trying to distinguish between the `Beamline().brho_ref` and the `Bunch().rigidity_ref`?

DavidSagan:
Another thought is `k_norm` since it is the normalization  value to convert between field and normalized bend strength.

mattsignorelli:
Would it really be that confusing to just tell people that we use a "signed" `rigidity`? `k_norm` almost feels more obscure to me

eiad-hamwi:
@ georg

Personally I don't think it's confusing to use a signed `rigidity` since `rigidity := p/q` but I haven't been working in the field for decades lol

DavidSagan:
One way to approach this would be to ask some people at Brookhaven about whether they would be confused by a brho with a sign.

mattsignorelli:
Closing this because it seems we will find a different solution

PR #110, opened 2025-06-18
Symplectic integration for the orbit

JosephPeterDevlin:
Note that I had to change Dan's multipole convention to match Bmad, so his tests are commented out for now. All methods have tests with zero curvature. For elements with curvature, we need to work on the SBend first. Note that the tolerances for some of my tests are suboptimal, and this is likely due to subtle differences in the integration steps between my code and PTC (for example, PTC does not include nonlinear pz dependence in the matrix part of matrix-kick).


mattsignorelli:
Wow! Very great work!

Dan has mentioned that we do not need to have these length 21 arrays for the multipoles. These will definitely slow down the code because they are too large to fit in registers and there are three of them at least it looks like. Can the code be refactored to instead only make these arrays be of length = number of multipoles in the bdict? You can get these using `keys(bdict)` and `values(bdict)`. You may need to wrap both of these in a `collect` if you are iterating over them

mattsignorelli:
Something to note is that i the future I imagine `bdict` will be modified to instead be internally represented as a static array for the keys and a static array for the value (where `Beamlines.jl` handles adding new multipoles to the element). Therefore in the future, `bdict.keys` will be a static array of length = n_multipoles, and `bdict.values` will be the multipoles. To emulate this behavior now, you can do `collect(keys(bdict))` and `collect(values(bdict))` which will create regular arrays. There will be allocations but that is ok right now - just do it in layer 2 and not inside of a kernel (layer 1)

ext/BeamTrackingBeamlinesExt/integration.jl
```diff
@@ -0,0 +1,187 @@
...
+@inline thin_pure_bmultipole(tm::Integration, bunch, bm1) = thin_pure_bdipole(tm, bunch, bm1)
+
+@inline thin_bmultipole(tm::Integration, bunch, bm1) = thin_bdipole(tm, bunch, bm1)
+
+
+# === Thick elements === #
+@inline function drift(tm::Integration, bunch, L)
+  tilde_m, gamsqr_0, beta_0 = ExactTracking.drift_params(bunch.species, bunch.Brho_ref)
+  return KernelCall(ExactTracking.exact_drift!, (beta_0, gamsqr_0, tilde_m, L))
+end
```
mattsignorelli:
```suggestion
@inline drift(tm::Integration, bunch, L) = drift(Exact(), bunch, L)
```

ext/BeamTrackingBeamlinesExt/integration.jl
```diff
@@ -0,0 +1,187 @@
...
+
+@inline function thick_pure_bsolenoid(tm::Integration, bunch, bm0, L) 
+  tilde_m, gamsqr_0, beta_0 = ExactTracking.drift_params(bunch.species, bunch.Brho_ref)
+  Ks = get_thick_strength(bm0, L, bunch.Brho_ref)
+  return KernelCall(ExactTracking.exact_solenoid!, (Ks, beta_0, gamsqr_0, tilde_m, L))
```
mattsignorelli:
```suggestion
@inline thick_pure_bsolenoid(tm::Integration, bunch, bm0, L) = thick_pure_solenoid(Exact(), bunch, bm0, L)
```

mattsignorelli:
Also I think that this should basically be `SciBmadStandard` tracking instead of `Integration`. Pinging @DavidSagan for thoughts

mattsignorelli:
Because what you've implemented here customizes the split based on the element . But maybe someone wants to do DKD through a pure quadrupole, so that option should still be there. Therefore this should be scibmadstandard, and we can split up what you've implemented into several tracking methods (matrix-kick, drift-kick, etc etc)

JosephPeterDevlin:
> Because what you've implemented here customizes the split based on the element . But maybe someone wants to do DKD through a pure quadrupole, so that option should still be there. Therefore this should be scibmadstandard, and we can split up what you've implemented into several tracking methods (matrix-kick, drift-kick, etc etc)

I agree with this, but maybe just a flag in the integration struct instead of an entirely different tracking method for each split. I did not add this yet because I was not sure of how best to implement a field with only a few discrete choices (maybe a string that throws an error if it's not one of the possible values?).

And also then SciBmadStandard can get my unpacking file and we can just make a new unpacking file for integration.

mattsignorelli:
> I was not sure of how best to implement a field with only a few discrete choices (maybe a string that throws an error if it's not one of the possible values?).

An `EnumX` is exactly what you want. See the `State` thing in `types.jl`.

> but maybe just a flag in the integration struct instead of an entirely different tracking method for each split. 

I'm not sure about this - there is no cost to a new struct, and by using a new struct invalid tracking methods will automatically throw an error with the generic unpack framework in place. Maybe we should discuss on a call

JosephPeterDevlin:
> I'm not sure about this - there is no cost to a new struct, and by using a new struct invalid tracking methods will automatically throw an error with the generic unpack framework in place. Maybe we should discuss on a call

Yeah, we can discuss. We also need to discuss if we want to allow both ds_step and n_steps.

JosephPeterDevlin:
@mattsignorelli I just pushed a new version with a lot of the changes we discussed yesterday (mostly to the tracking method structs). note that I had to name it Standard (not SciBmadStandard) for now because of a name conflict with what I believe is a placeholder from Beamlines.

JosephPeterDevlin:
@mattsignorelli I think it is ready to go. Note that I have commented out Dan's tests because I modified the parameter list for his functions slightly. If we think these tests are necessary, I can add them back, but I believe they are made obsolete by my tests.

mattsignorelli:
@JosephPeterDevlin Amazing. 

We should keep Dan's tests in there, since he spent time on them and they are good checks for nonlinear stuff. Could you add them back? And resolve the merge conflict too

JosephPeterDevlin:
@mattsignorelli I just recalled a bigger problem for Dan's matrix-kick tests. Dan's function was originally matrix-kick-matrix, but since Etienne uses kick-matrix-kick in PTC, I had to change the order to check against PTC. I can add back the other tests, though.

ext/BeamTrackingBeamlinesExt/integration.jl
```diff
@@ -0,0 +1,167 @@
...
+  params = (beta_0, brho_0, 0, b0, 0, 0, mm, kn, sn)
+  return integration_launcher!(IntegrationTracking.bkb_multipole!, params, tm, L)
+end
+=#
+@inline function thick_pure_bquadrupole(tm::Union{SplitIntegration,MatrixKick}, bunch, bm, L)
+  brho_0 = bunch.Brho_ref
+  tilde_m, gamsqr_0, beta_0 = ExactTracking.drift_params(bunch.species, brho_0)
+  mm = bm.order
+  kn, ks = get_strengths(bm, L, brho_0)
+  kn = @MArray [kn]
```
mattsignorelli: This will be slow. What we should do is have an integer argument in `multipole_kick!` that ignores any multipoles of order <= integer argument. Then we don't have to allocate arrays like this

ext/BeamTrackingBeamlinesExt/integration.jl
```diff
@@ -0,0 +1,167 @@
...
+  mm, kn, sn = get_thick_multipoles_no_dipole(bunch, bm, L)
+  params = (beta_0, brho_0, 0, b0, 0, 0, mm, kn, sn)
+  return integration_launcher!(IntegrationTracking.bkb_multipole!, params, tm, L)
+end
+=#
+@inline function thick_pure_bquadrupole(tm::Union{SplitIntegration,MatrixKick}, bunch, bm, L)
+  brho_0 = bunch.Brho_ref
+  tilde_m, gamsqr_0, beta_0 = ExactTracking.drift_params(bunch.species, brho_0)
+  mm = bm.order
+  kn, ks = get_strengths(bm, L, brho_0)
```
mattsignorelli, line 93: It looks to me like we do not support this tracking method with any skew quadrupole term yet, so we will need either an error or a call to `patch!` to patch in and patch out of the plane where the quadrupole term is only "normal".

ext/BeamTrackingBeamlinesExt/integration.jl
```diff
@@ -0,0 +1,167 @@
...
+  params = (beta_0, gamsqr_0, tilde_m, mm, kn, ks)
+  return integration_launcher!(IntegrationTracking.dkd_multipole!, params, tm, L)
+end
+
+@inline function thick_bquadrupole(tm::Union{SplitIntegration,MatrixKick}, bunch, bm, L)
+  brho_0 = bunch.Brho_ref
+  tilde_m, gamsqr_0, beta_0 = ExactTracking.drift_params(bunch.species, brho_0)
+  mm = bm.order
+  kn, ks = get_strengths(bm, L, brho_0)
```
mattsignorelli: likewise skew quad term not supported yet

src/modules/IntegrationTracking.jl
```diff
@@ -0,0 +1,326 @@
...
+tilde_m:  1 / (βγ)_0  # mc^2 / p0·c
+mm: vector of m values for non-zero multipole coefficients
+kn: vector of normal multipole strengths scaled by Bρ0
+ks: vector of skew multipole strengths scaled by Bρ0
+L: element length
+"""
+@makekernel fastgtpsa=true function mkm_quadrupole!(i, b::BunchView, beta_0, gamsqr_0, tilde_m, mm, kn, ks, L)
+  k1 = kn[1]
+  kn_new = copy(kn)
+  kn_new[1] = 0
```
mattsignorelli: This will be super slow, see the recommendation above to add another argument to `multipole_kick!`

mattsignorelli (changes requested):
Amazing work. Just a couple of minor points. I think we should try to support skew quads via a patch in and patch out addition to the KernelChain. let me know what you think.

Also we need tests for allocations inside of the low-level kernels. there should be zero so we have GPU compatibility

ext/BeamTrackingBeamlinesExt/integration.jl
JosephPeterDevlin: Skew quads are supported, but they are included in the multipole kick and not the matrix part. Either way I agree that the patch-in-patch-out is a better solution.

ext/BeamTrackingBeamlinesExt/integration.jl
JosephPeterDevlin: Agreed

src/modules/IntegrationTracking.jl
JosephPeterDevlin: Agreed, will be fixed by the patch-in-patch-out for matrix part

ext/BeamTrackingBeamlinesExt/integration.jl
JosephPeterDevlin: See above

JosephPeterDevlin:
> Also we need tests for allocations inside of the low-level kernels. there should be zero so we have GPU compatibility

I tested this by hand, and I saw 2 or 3 more allocations than linear (but it did not grow with more steps). This may be due to the arrays I was allocating, although I'm not sure. Is there a systematic way to put this kind of test in?

JosephPeterDevlin:
@mattsignorelli I believe I have addressed all the requested changes

mattsignorelli:
> I tested this by hand, and I saw 2 or 3 more allocations than linear (but it did not grow with more steps). This may be due to the arrays I was allocating, although I'm not sure. Is there a systematic way to put this kind of test in?

See the function `test_map` in `runtests.jl`. Tests for `ExactTracking` use this function - which goes straight to the Kernel and does not know anything about Beamlines.

You should keep your `Beamlines` tests in, because they are needed. But add test_map calls for your functions . These should probably go in a new file `splitintegration.jl` in the test suite

JosephPeterDevlin:
@mattsignorelli I added some tests for the kernels directly, but upon further inspection, there are allocations. The kernel with allocations simply calls two other kernels, each of which has been individually verified to have no allocations, so the allocations must come from calling a kernel within a kernel. We will have to discuss this.

JosephPeterDevlin:
@mattsignorelli New version with no allocations is in.

ext/BeamTrackingBeamlinesExt/integration.jl
```diff
@@ -0,0 +1,171 @@
...
+  Ksol = sqrt(kn[1]^2 + ks[1]^2)
+  params = (beta_0, gamsqr_0, tilde_m, Ksol, mm, kn, ks)
+  return integration_launcher!(IntegrationTracking.sks_multipole!, params, tm, L)
+end
+
+@inline function thick_pure_bdipole(tm::Union{SplitIntegration,DriftKick}, bunch, bm, L)
+  brho_0 = bunch.Brho_ref
+  tilde_m, gamsqr_0, beta_0 = ExactTracking.drift_params(bunch.species, brho_0)
+  mm = bm.order
+  kn, ks = get_strengths(bm, L, brho_0)
```
mattsignorelli, line 64:
```suggestion
  knt, kst = get_strengths(bm, L, brho_0)
  kn = SVector(knt)
  ks = SVector(kst)
```

Instead of multiplying L inside the kernel, let's just make `kn` and `ks` be `SVector`s (stack allocated) instead of `SizedVector`s.

mattsignorelli:
@JosephPeterDevlin  I made a change to the `get_strengths` functions which makes it so we can use [57723dc](https://github.com/bmad-sim/BeamTracking.jl/pull/110/commits/57723dcfb05fe60d0a90940fa8fbb6378b8fd175). I branched this specific commit with the update in #119  and got re-changed `start` to a positional argument.

mattsignorelli:
PR superceded by #119  .

PR #111, opened 2025-06-18
Exact Bend with tests

Alexheeeee:
- implemented exact bend 
- includes matrix kernel test and BeamlinesExt test
- 10th order map test disagrees with PTC (commented out for now)
- accepts negative length
- thick bend and straight pure dipole implemented for exact tracking

P.S. BitsBeamline is crashing because of GTPSA error


src/utils.jl
```diff
@@ -130,12 +130,12 @@ massof(s::Species) = s.mass
...
 # Particle energy conversions =============================================================
-calc_Brho(species::Species, E) = @FastGTPSA sqrt(E^2-massof(species)^2)/C_LIGHT/chargeof(species)
+calc_Brho(species::Species, E) = @FastGTPSA sqrt(E^2-massof(species)^2)/C_LIGHT/abs(chargeof(species))
 calc_E(species::Species, Brho) = @FastGTPSA sqrt((Brho*C_LIGHT*chargeof(species))^2 + massof(species)^2)
 calc_gamma(species::Species, Brho) = @FastGTPSA sqrt((Brho*C_LIGHT/massof(species))^2+1)
 
-calc_p0c(species::Species, Brho) = @FastGTPSA Brho*C_LIGHT*chargeof(species)
-calc_beta_gammma(species::Species, Brho) = @FastGTPSA Brho*chargeof(species)*C_LIGHT/massof(species)
+calc_p0c(species::Species, Brho) = @FastGTPSA Brho*C_LIGHT*abs(chargeof(species))
+calc_beta_gammma(species::Species, Brho) = @FastGTPSA Brho*abs(chargeof(species))*C_LIGHT/massof(species)
```
mattsignorelli: If you allow a negative Brho for negative charge (p/q) then wouldn't everything still be consistent? I think we should leave this for now . I have some ideas we can all discuss

mattsignorelli:
The test error message `error: /workspace/srcdir/gtpsa/code/mad_tpsa_fun.c:216: : /workspace/srcdir/gtpsa/code/mad_tpsa_fun.c:216: ` in GTPSA corresponds to this line: https://github.com/mattsignorelli/gtpsa/blob/6147f1ce0977cfcb79c3225108c32594faad9047/code/mad_tpsa_fun.c#L216

Basically the code is taking a sqrt of a nonpositive number somewhere. But the derivative does not exist here. 

I just saw your comment: 
> P.S. BitsBeamline is crashing because of GTPSA error

Did you add tests with BitsBeamline? There is currently no error in the main branch

mattsignorelli:
> 10th order map test disagrees with PTC (commented out for now)

Also David has the exact bend coded in Bmad so they should agree exactly

eiad-hamwi:
> > 10th order map test disagrees with PTC (commented out for now)
> 
> Also David has the exact bend coded in Bmad so they should agree exactly

From what I can tell, the reason it doesn't agree is because Bmad asks PTC to calculate the higher-order TPS map (anything nonlinear), and then PTC defaults to "drift-kick" integration of bends when generating the TPS map. 

So maybe one way we can try to squeeze through this is by changing the `integration_order` to like 8 or something, and increasing the number of steps to 1000s or something, so we hit machine precision? I still think this will probably not agree super well at higher-orders of the map because it's just a fundamentally different way of tracking through the bend. 

And even if we can somehow get the TPS map from PTC for the PTC implementation of the exact bend, then the higher orders may still not agree because there is no PTC implementation for Bmad's exact bend (which has no singularities).

What is nice currently is that the jacobian (generated from Bmad's own exact bend) agrees to all shown decimals with this implementation, and also the tracking of some random initial conditions also agrees to all shown decimals. So I think this actually is an equivalent way of proving the full nonlinear agreement if we just compare a few different initial conditions.

mattsignorelli:
@DavidSagan is there any way to get a map from Bmad for your exact bend?

Nonetheless, something is wrong here -- GTPSA's error says something is not differentiable, because it is evaluating a derivative at the bound of some function. I don't see any added code using `BitsBeamline`, and the current CI for `BitsBeamline` works just fine, so I don't see how it could be related to that.

Also even so, we need more decimal places for the matrix comparison. `e23.16` I think is the fortran number formatter you want. See the options in Tao `help show taylor` to printhigher precision

src/utils.jl
eiad-hamwi: @Alexheeeee you should keep the `abs()` in `calc_p0c` and `calc_beta_gamma` around the charge term, since Matt was only suggesting to remove it from `calc_Brho`.

@mattsignorelli, should Alex also separately add an `abs()` around the `Brho` term in `calc_p0c` and `calc_beta_gamma`, since people may want to pass a negative value of `Brho` but still get a positive result?

mattsignorelli: If you just keep the original definitions but rightfully require negative Brho for negative charge, then the signs will all handle themselves, so no `abs` needed

eiad-hamwi: Yep, that's true! My bad.

eiad-hamwi:
The bug is occurring because there is a constraint in the `sqrt` that should be enforced: `cplus^2 + gp * alpha > 0`

This is equivalent to requiring that 
```
| k0 / pt * (1 + gx) * L * sinc(g * L)  - sin(g * L + asin(px / pt)) | < 1
```

This can be simplified for the reference trajectory `x = y = px = py = δ = 0` such that

```
g - 1 / (L sinc(g L)) < k0 < g + 1 / (L sinc(g L))
```
The most restrictive case happens when `g = 0`, which reduces further to:
```
- 1/L < k0 < 1/L
```
In the tests we were using, we chose `g = 0` and `K0L = 1`, which clearly violates this inequality. 

If we consider the more general coordinate-dependent case, and assume that `g = 0` is still the most restrictive scenario in the general case, we find that:

```
(k0 L - px)^2 + py^2 < (1+pz)^2
```
I haven't considered yet the physical interpretation of this condition. It almost seems like it's some paraxial condition, which should not be the case in this function.

On the one hand, we can add a simple conditional statement at the `BeamlinesExt` level to avoid this scenario for the zero trajectory. But on the other hand, this may also crash in a coordinate-dependent way, and we would ideally address this at the kernel level as well. I'm not sure on how severe the coordinate-dependence is, however, since this whole situation is a very minor edge case anyway.

eiad-hamwi:
@Alexheeeee replace `K0` with `Kn0` on line 854 in your `test/ExactTracking.jl` file to match the new parameter names.

ext/BeamTrackingBeamlinesExt/exact.jl
```diff
@@ -14,4 +14,18 @@ end
 @inline function drift(tm::Exact, bunch, L)
   tilde_m, gamsqr_0, beta_0 = ExactTracking.drift_params(bunch.species, bunch.Brho_ref)
   return KernelCall(ExactTracking.exact_drift!, (beta_0, gamsqr_0, tilde_m, L))
+end
+
+@inline function thick_bend_pure_bdipole(tm::Exact, bunch, bendparams, bm1, L)
+  g = bendparams.g
+  theta = g * L
+  k0, _ = get_strengths(bm1, L, bunch.Brho_ref)
```
mattsignorelli:
```suggestion
  Kn0, Ks0 = get_strengths(bm1, L, bunch.Brho_ref)
```

And we need to be able to support nonzero `Ks0`. This probably can be done using a patch to rotate into a plane where you only have a `Kn0` and then rotate back.  If someone commits to implementing this soon then I will merge this but with add the error message:
```suggestion
  Kn0, Ks0 = get_strengths(bm1, L, bunch.Brho_ref)
  Ks0 ≈ 0 || error("Skew/tilt multipoles not implemented yet for tracking method $tm")
```

src/modules/ExactTracking.jl
```diff
@@ -326,6 +326,50 @@ to carry both reference and design values.
...
+  h = 1+g*v[i,XI] # 1 + g * x
+  cplus = cos(phi1) #cos(theta + phi1)
+  sinc_theta = sincu(theta) #sincu(theta)
+  alpha = 2*h*sin(phi1)*abs(L)*sinc_theta- gp*h^2*L^2*(sinc_theta)^2 #alpha
+  cond = cplus^2 + gp*alpha
+  #particle does not intersect the exit face
+  if cond <= 0
+    b.state[i] = State.Lost
+    return
+  end
```
mattsignorelli: We should not create a branch like this because it kills SIMD. What we should do is instead compute both and let a *0 or *1 define if the particle is propagated. For a simple example the drift:

```julia
alive = ifelse(b.state[i]==State.Alive, 0, 1) # should be branchless
v[i,XI] = alive*(v[i,XI] + L*v[i,PXI]) - (alive - 1)*v[i,XI] # branchless drift

# This MAY work too but not guaranteed according to the documentation
v[i,XI] = ifelse(alive, v[i,XI] + L*v[i,PXI], v[i,XI]) 
```

Here you see if alive == 1, then the particle is drifted. If it is 0, then it is kept the same

For the bend where there is a singularity if the particle dies, you can include a value for the "alive" part inside the square root to maintain differentiability. E.g.

```julia
b.state[i] = ifelse(cond <= 0 && b.state[i] == State.Alive, State.Lost, b.state[i]) # should be branchless
alive = ifelse(b.state[i]==State.Alive, 0, 1) # should be branchless
nasty_sqrt = alive*sqrt(cond + (alive-1)*(1+cond))
```

See here now if alive == 1 then the sqrt is taken normally, but if it is 0, then the sqrt will be `0*sqrt(cond - 1 - cond) = 0*sqrt(1)` which is safe and differentiable (no sqrts of 0) and branchless.

src/modules/ExactTracking.jl
```diff
@@ -326,6 +326,50 @@ to carry both reference and design values.
...
+  Lcv = -L*sinc_theta-sign(L)*v[i,XI]*sin(theta) #Lcv
+  thetap = 2 * (phi1 - sign(L)*atan(xi, -Lcv)) #theta_p
+  Lp = sign(L)*sqrt(Lcv^2 + xi^2) / sincu(thetap/2) #Lp
+
+  v[i,XI] = v[i,XI]*cos(theta) - g/2*L^2*(sincu(theta/2))^2 + xi
+  v[i,PXI] = pt*sin(phi1 - thetap)
+  v[i,YI] = v[i,YI] + v[i,PYI]*Lp/pt 
+  v[i,ZI] = v[i,ZI] - rel_p*Lp/pt + 
+                  abs(L)*rel_p/sqrt(tilde_m^2+rel_p^2)/beta_0
```
mattsignorelli: add in the `alive` part as I showed previously with the drift example so that particles are only propagated when alive

test/ExactTracking.jl
```diff
@@ -758,8 +758,108 @@ zf_mn4  = [ 0., 3.140908277834687e-8, -3.1503450227072763e-8, 3.140908186274627e
...
+    Brho_ref = BeamTracking.calc_Brho(ELECTRON, sqrt(p0c^2 + BeamTracking.massof(ELECTRON)^2))
+    ele_bend = LineElement(L=0.5, g=-1, Kn0 = -0.9, tracking_method=Exact())
+    ps = [0.1, -7.5e-4, 0.1, -3e-4 , 0.1, -1e-3]
+    b0 = Bunch(collect(transpose(@vars(D1) + ps)), Brho_ref=Brho_ref)
+    bl = Beamline([ele_bend], Brho_ref=Brho_ref)
+    track!(b0, bl)
+
+    @test GTPSA.jacobian(b0.v) ≈ exact_bend_7
+  end
+
```
mattsignorelli, line 847: Because we cannot test the higher order map, we do need tests for individual scalar particles, ideally with large amplitudes.

mattsignorelli (changes requested):
Thanks for your work on this! A couple of changes:
1. Explicitly state lack of support for vertical bends, or add in the patch (someone can help you maybe?)
2. Branchless programming for state setting
3. Add tests for individual particles with large amplitudes

ext/BeamTrackingBeamlinesExt/exact.jl
JosephPeterDevlin: @mattsignorelli I will implement the patching if this gets pulled in. I can also add some more tests/make it branchless if necessary.

mattsignorelli: @JosephPeterDevlin you have write permission in this repository, so you could actually pull this branch, work on it, and then push back to this branch

mattsignorelli:
@Alexheeeee could you respond to each of the points above and "resolve" the conversation if you resolved it? Thanks!

test/ExactTracking.jl
Alexheeeee: I added one test, many more needed

ext/BeamTrackingBeamlinesExt/exact.jl
Alexheeeee: I haven't worked on this yet

JosephPeterDevlin: @mattsignorelli The BendParams currently only has one number for g. I assume this means the curvature in the horizontal plane, but that means we cannot currently support vertical bends, and only what you would call a vertical corrector/kicker.

DavidSagan: `tilt_ref` rotates the bend so you can get a vertical bend.

JosephPeterDevlin: What is tilt_ref? Do you mean the alignment parameter? If so, that's probably true, but that should most likely be handled when misalignments get included.

JosephPeterDevlin: Oh, I assume you mean ref_tilt from classic Bmad. Does that exist in SciBmad?

DavidSagan: Correct.

mattsignorelli: In `SciBmad` we could have the curvature values `g_x` and `g_y` instead, and I guess optionally another independent variable `g_tilt`. This is all for the curvature of the coordinate system - irrespective of the bend field strength, which is specified by `Kn0`, `Ks0`, and `tilt0`

JosephPeterDevlin: Yeah, the patches are not so simple then. If I'm understanding correctly, the exact bend could only be patched into if `Ks0 = 0` for any `g_tilt` and `gy = 0` or if `atan(gy/gx) = atan(Ks0/Kn0)` with `g_tilt = 0`.

DavidSagan: `gx` and `gy` are possible but `tilt_ref` is more flexible. I don't understand what the difference is  between `g_tilt` and `tilt_ref`

mattsignorelli: No difference, I just think `g_tilt` is more clear than `tilt_ref` personally.

@JosephPeterDevlin For nonzero `g_tilt` one would just decompose that into solely `g_x` and `g_y`, just like we do with the multipoles in `get_strengths`. Then after that just use `g_x` and `g_y`.

DavidSagan: Also keep in mind that `tilt` is different from `tilt_ref` (and that no other type of element has a `tilt_ref`)

JosephPeterDevlin: @mattsignorelli What I mean is that the bend Alex implemented only works if `g` is in the same plane as the dipole field. A general solution for curvature in any plane might exist, but I've never seen it.

DavidSagan: Actually with `tilt_ref` the reference curve is being tilted. In both cases, `tilt` and `tilt_ref` will tilt `g` so `g_tilt` is a bit ambiguous.

DavidSagan: @JosephPeterDevlin Remember that the tracking is done in body coordinates and in body coordinates the bend rotation is always around the `y`-axis independent of `tilt_ref`. So Alex's implementation should work.

mattsignorelli: @DavidSagan @JosephPeterDevlin can we meet on zoom to discuss?

DavidSagan: Sure. Go to my zoom.

JosephPeterDevlin: I'm available.

mattsignorelli (approved):

PR #149, opened 2025-08-14
Spin (with testing)

JosephPeterDevlin:
Notes:

- I switched patches to use quaternions instead of rotation matrices. This saves a big headache for spin through a patch and only affects internals, not user experience.
- A "fake APC" is being used for (g-2)/2 because of some issues for now. This will be easy to fix later.
- Although TPSA spin is pretty slow (note the test times now), scalar spin seems to have very little slowdown compared to purely orbital tracking.


mattsignorelli:
Amazing!!

mattsignorelli (approved):

mattsignorelli:
Is the TPSA spin tracking slower than PTC?

JosephPeterDevlin:
> Is the TPSA spin tracking slower than PTC?

What I noticed is that the TPSA spin in SciBmad becomes prohibitively slow at high order (10) with many steps (like an 8th order integrator). Otherwise it is fine (for a sixth-order integrator and 1st order map, it seems to be only about 4x slower than purely orbital tracking for the ESR). This may be a memory problem since I was running PTC on the CLASSE cluster and SciBmad on my laptop.

mattsignorelli:
What about say a 3rd order map? How much slower than orbital?

I think a `@FastGTPSA` block wrapped around the calls in `quat_mult` could improve the performance a bit but not by x4

JosephPeterDevlin:
> What about say a 3rd order map? How much slower than orbital?
> 
> I think a `@FastGTPSA` block wrapped around the calls in `quat_mult` could improve the performance a bit but not by x4

About 6x slower than pure orbital at third order

JosephPeterDevlin:
In case you're curious, only about 20% slower for Float64

mattsignorelli:
Can you modify quat miult so:

```julia
@inline function quat_mul(q1, q2)
  """
  Returns q1 * q2.
  """
  a1, b1, c1, d1 = q1[Q0], q1[QX], q1[QY], q1[QZ]
  a2, b2, c2, d2 = q2[Q0], q2[QX], q2[QY], q2[QZ]
  @FastGTPSA begin
    a3 = a1*a2 - b1*b2 - c1*c2 - d1*d2
    b3 = a1*b2 + b1*a2 + c1*d2 - d1*c2
    c3 = a1*c2 - b1*d2 + c1*a2 + d1*b2
    d3 = a1*d2 + b1*c2 - c1*b2 + d1*a2
  end
  return SA[a3 b3 c3 d3]
end
```

How much worse then?

mattsignorelli:
Also modify `rotate_spin!` so:

```julia
@makekernel fastgtpsa=false function rotate_spin!(i, coords::Coords, a, g, tilde_m, mm, kn, ks, L)
  """
  This function rotates particle i's quaternion according to the multipoles present.
  """
  q2 = coords.q
  q1 = expq(-L/2 * omega(i, coords, a, g, tilde_m, mm, kn, ks))
  q3 = quat_mul(q1, SA[q2[i,Q0], q2[i,Q1], q2[i,Q2], q2[i,Q3]])
  q2[i,Q0], q2[i,QX], q2[i,QY], q2[i,QZ] = q3[Q0], q3[QX], q3[QY], q3[QZ]
end
```

mattsignorelli:
this is because `q2[i,:]` is a allocates a new array so this would actually not work on the GPU too

JosephPeterDevlin:
Now about 3.5x slower for 1st order and 4.8x slower for 3rd order. I noticed that the memory allocation grows in proportion with the time, so I believe my suspicion is correct.

mattsignorelli:
So we have to find ways to reduce allocations. One way is to pre-allocate the TPSAs at the beginning. E.g.

```julia
a = zero(v[i,XI])  # will be stack allocated for regular numbers
b = zero(v[i,XI])  # TPSA allocated for TPSAs

@FastGTPSA! v[i,XI] = a + b # zero allocations, works for both
```

JosephPeterDevlin:
Before we get there, I think some strategic FastGTPSAs are also in order. For example, this alone saved 3% of the allocations:
```
  @FastGTPSA begin
  # kinetic momenta, not canonical momenta
    px = v[i,PXI] + (v[i,YI] * kn[1] / 2) * (mm[1] == 0) 
    py = v[i,PYI] - (v[i,XI] * kn[1] / 2) * (mm[1] == 0)
  end
```

mattsignorelli:
can you try setting the kwarg `fastgtpsa=true` for `@makekernel` or is there still the `isless` bug?

JosephPeterDevlin:
This is in omega, which is not a kernel because it returns the precession vector. I wrapped the whole function in FastGTPSA, though, and that saves another 5% of the allocations

JosephPeterDevlin:
I wrapped everything that seems to make a difference in @FastGTPSA and at 3rd order that brought the memory down about 16%. I think there is more to be done, but I'm not sure exactly what. Maybe preallocations somewhere.

JosephPeterDevlin:
Discussion should be moved to #150

PR #150, opened 2025-08-14
Spin optimizations

JosephPeterDevlin:


JosephPeterDevlin:
Some notes: it is possible to reduce the allocations by almost 50% through preallocating TPSs and placing `@FastGTPSA!` everywhere. However, this leads to errors in many cases, including type instabilities or flat out wrong numbers. I think this needs to be fixed by systematically going through the code with a GTPSA expert (@mattsignorelli) and carefully preallocating/placing macros.

ext/BeamTrackingBeamlinesExt/integration.jl
mattsignorelli, line 1: Do the `@inbounds` here have any effect? I know thye are necessary inside the kernel for SIMD but here it might be worth keeping the bounds check in

src/modules/ExactTracking.jl
```diff
@@ -106,33 +106,35 @@ properties, though I've not seen a proof of that claim.
...
+      m -= 1
+      t  = (by * x - bx * y) / m
+      bx = (by * y + bx * x) / m
+      by = t
+      add = (0 < jm && m == ms[jm]) && (m != excluding) # branchless
+      idx = max(1, jm) # branchless trickery
+      by += knl[idx] * add
+      bx += ksl[idx] * add
+      jm -= add
+    end end
```
mattsignorelli, line 137:
```suggestion
  @FastGTPSA begin
    jm = length(ms)
    m  = ms[jm]
    add = (m != excluding && m > 0)
    by = knl[jm] * add
    bx = ksl[jm] * add
    jm -= 1
    while 2 <= m
      m -= 1
      t  = (by * x - bx * y) / m
      bx = (by * y + bx * x) / m
      by = t
      add = (0 < jm && m == ms[jm]) && (m != excluding) # branchless
      idx = max(1, jm) # branchless trickery
      by += knl[idx] * add
      bx += ksl[idx] * add
      jm -= add
    end
```

Note that `@makekernel` automatically wraps everything in an `@inbounds`, which is then propagated into calling functions. So here we do not need a `@inbounds`

src/modules/IntegrationTracking.jl
```diff
@@ -387,53 +387,74 @@ end
...
   """
   q2 = coords.q
-  q1 = expq(-L/2 * omega(i, coords, a, g, tilde_m, mm, kn, ks))
-  q3 = quat_mul(q1, q2[i,:])
-  q2[i,Q0], q2[i,QX], q2[i,QY], q2[i,QZ] = q3[Q0], q3[QX], q3[QY], q3[QZ]
+  q1 = expq(-omega(i, coords, a, g, tilde_m, mm, kn, ks, L)/2)
+  @inbounds begin
+    q3 = quat_mul(q1, SA[q2[i,Q0], q2[i,QX], q2[i,QY], q2[i,QZ]])
+    q2[i,Q0], q2[i,QX], q2[i,QY], q2[i,QZ] = q3[Q0], q3[QX], q3[QY], q3[QZ]
+  end
```
mattsignorelli, line 457:
```suggestion
  q3 = quat_mul(q1, SA[q2[i,Q0], q2[i,QX], q2[i,QY], q2[i,QZ]])
  q2[i,Q0], q2[i,QX], q2[i,QY], q2[i,QZ] = q3[Q0], q3[QX], q3[QY], q3[QZ]
```

`@makekernel` includes automatically a `@inbounds`

src/utils.jl
```diff
@@ -93,86 +93,102 @@ function sincuc(x)
...
+  """
+  ε = eps(T)
+  N_max = 100
+  N = 1
+  conv = false
+  y = one(x)
+  prev = one(x)
+  result = one(x)
+  @FastGTPSA! begin
+    if x < 0.1
```
mattsignorelli, line 110: This branch here will kill the SIMD. is there a way we can get rid of it?

src/utils.jl
mattsignorelli, line 1: We should also have a function like `sincos_quaternion` which computes both together using only one loop for speed, if iteration is necessary

mattsignorelli: Also, what exactly is the `sin(sqrt(x))` where `x` is a negative number? doesn't that return a complex number?

mattsignorelli (changes requested):

src/modules/ExactTracking.jl
JosephPeterDevlin: Doesn't that apply to FastGTPSA too then?

src/utils.jl
JosephPeterDevlin: Putting them together is a good idea. This function is only called for x >= 0, it doesn't make sense for negative x.

src/utils.jl
JosephPeterDevlin: I borrowed this implementation from PTC. Is there another way to define it with the correct derivatives around zero? The problem is that `sqrt(x)` does not have a Taylor expansion around zero, but `sinc(sqrt(x))` does.

Could this be fixed by implementing the function directly in `GTPSA.jl`?

ext/BeamTrackingBeamlinesExt/integration.jl
JosephPeterDevlin: I can get rid of them

src/utils.jl
JosephPeterDevlin: Actually, does something like this work? A little slower but technically branchless.
```
function sin_quaternion(x::T) where T
  """
  This function computes sin(sqrt(x))/sqrt(x), which is necessary for
  exponentiating a rotation vector into a quaternion.
  """
  ε = eps(T)
  N_max = 100
  N = 1
  conv = false
  small = (x < 0.1)
  y = one(x)
  prev = one(x)
  result = one(x)
  z = one(x)
  @FastGTPSA! begin
    while small && !conv && N < N_max
      y = -y*x/((2*N + 1)*(2*N))
      result = prev + y
      N += 1
      if norm_tps(result - prev) < ε
        conv = true
      end
      prev = result
    end
    z = ifelse(small, z, x)
    result = ifelse(small, result, sin(sqrt(z))/sqrt(z))
  end
  if N == N_max
    @warn "sin_quaternion convergence not reached in $N_max iterations"
  end
  return result
end
```

JosephPeterDevlin: When using this new implementation, all the TPSA buffers run out of temporaries. Is there a way to fix this other than splitting up expressions everywhere? It seems a bit odd that this one change causes that.

mattsignorelli: This still seems to have a branch to me, because the while loop can terminate at different iterations for different `x`, right? 

Honestly what we should do to at least check for branches is go back to explicit SIMD using `SIMD.jl`. There is a bug with that https://github.com/eschnett/SIMD.jl/issues/140 but we should just find a way to work around it or use functions with `fastgtpsa=false` (bug only shows with `fastgtpsa=true`). the `use_explicit_SIMD` flag can be passed to `track!` or `launch!`. But honestly, given how stupid Julia's compiler can be at times with this autovectorization, I would support making `use_explicit_SIMD=true` be the default again on the CPU.

As for the TPSA buffer, that must be a bug with the `@FastGTPSA!` block parsing. I would have to look at the `@macro_expand` code to understand what is happening. For now we can just leave it out.

src/utils.jl
mattsignorelli: OK so you use the power series expansion for small `x` basically. Instead of iterating, why not just use the power series expansion up to a certain order? E.g. like `sincu` does and other functions

mattsignorelli: actually why do we even need a power series for small x with `sin(sqrt(x))`? both `sqrt(x)` and `sin(x)` are well behaved near 0

src/utils.jl
JosephPeterDevlin: If a while-loop is not possible, that would indeed be pretty restrictive. I'm not sure if that counts as a branch.

src/utils.jl
JosephPeterDevlin: `sqrt(x)` is not differentiable at zero because its slope tends towards infinity. `sin(sqrt(x))` has a finite derivative at zero.

Note that `sincu` has a branch. The problem with not iterating is that it's difficult, if not impossible, to know beforehand how many iterations to use. With a float, you can find out how many iterations are needed to reach machine precision for a small enough number, but for a TPSA, you have to fill in all the higher derivatives which usually takes more iterations than the scalar part indicates.

mattsignorelli: > Note that sincu has a branch

Excellent point. We should fix this. I really think we need to get this entire package working with explicit SIMD. That'll easily give 2-4x improvement on the CPU for any number of particles. For TPSA it will make zero difference.

> sqrt(x) is not differentiable at zero because its slope tends towards infinity. sin(sqrt(x)) has a finite derivative at zero.

I might be confused. derivative of `sin(sqrt(x))` is `cos(sqrt(x))/(2*sqrt(x))` which is infinity at `x=0` ?

JosephPeterDevlin: Sorry, meant to say `sinc(sqrt(x))` has a finite derivative at zero, which is the function I'm evaluating. It's the same for `cos(sqrt(x))`. If you evaluate them naively, you will get a divide-by-zero, but they should have a finite limit at zero.

It should be possible to write these functions properly (branchless) for floats. However, for TPSA, I'm not sure that it's possible for the reasons I stated above. Would it help if we defined these functions in `GTPSA.jl` instead? Then the iteration could happen inside that package instead. I think there must be some iteration in there already, as that's somewhat necessary for evaluating special functions.

mattsignorelli: yes that is exactly what we would do, just like how `sin`, `sinc`, and friends are defined in `GTPSA` for the `TPS` type specifically. We would follow the implementations in the C code which use `for` loops and take into account the max order for best performance.

Nonetheless, do you have any literature for this function? I am having trouble finding anything on it online

JosephPeterDevlin: I haven't found any literature on it. It came up naturally when I was trying to exponentiate omega into a quaternion, and the only other place I saw it mentioned was in the PTC code.

JosephPeterDevlin:
Closing because development is going in a different direction (SIMD), will return to it later

PR #155, opened 2025-08-28
Alignment and aperture devel.

DavidSagan:
Posting PR to make sure code devel is on the right track.


DavidSagan:
@mattsignorelli Please take a look.

mattsignorelli:
Looks mostly good, just some things:
- Alignment and aperture kernels should be added to the `ExactTracking` module, not given their own tracking methods, because they are "exact"
- Kernel call syntax is `KernelCall(kernel, args)`, e.g. `KernelCall(linear_drift!, (L, r56))`
- The function `quat_rotate!` mutates an array in place, this cannot be done on the GPU. instead use static arrays so it returns a 3-vector with the rotation. Also joe has some code for similar things already with spin, could that be used? 
- Likewise I see some array allocation inside a kernel - this cannot be done
- minor point but your new tree double checks for `PatchParams`, this could be optimized but minor point

Also note that this will make us have to increase the "length" of the `KernelChain` to 5, because now an element can call aperture at both ends (2), alignment at both ends (2), and one tracking kernel call (1). So the line `kc = KernelChain(Val{1}())` will have to be `kc = KernelChain(Val{5}())`

JosephPeterDevlin:
> * The function `quat_rotate!` mutates an array in place, this cannot be done on the GPU. instead use static arrays so it returns a 3-vector with the rotation. Also joe has some code for similar things already with spin, could that be used?

This is basically done already in the patch functions where I define the 3 x 3-matrix elements in terms of the quaternion elements. The code could be moved to a utility function.

mattsignorelli:
Hi David, I made a bunch of edits, please review

1. Changed tree structure so whenever `aperture` is called, it is active
2. Tree structure minimizes number of checks in common cases
3. `entering` always a boolean including for `alignment`
4. simplified structure of `aperture` funciton in the extension, as well as the `aperture!` kernel in the package core, ensuring SIMD compatibility
5.  fixed names of some variables

Let me know and also pinging @JosephPeterDevlin  for a review

DavidSagan:
@mattsignorelli One quick note is that there should not be multiple files with the same name and now there are two `alignment.jl` and `aperture.jl`.

mattsignorelli:
One thing is that the aperture kernel in the core package should have its own tests, separate from those tests which access the kernels at the `BeamlinesExt` level - those tests should go in `BeamlinesExt.jl` test file. Can you add these?

mattsignorelli:
> @mattsignorelli One quick note is that there should not be multiple files with the same name and now there are two `alignment.jl` and `aperture.jl`.

That can be changed, but `ApertureTracking.jl` is the convention used if the file contains a module named `ApertureTracking`, which is not true. Maybe `aperture_tracking.jl`?

DavidSagan:
OK any other name is fine with me.

DavidSagan:
```
One thing is that the aperture kernel in the core package should have its own tests, separate from those tests which access the kernels at the BeamlinesExt level - those tests should go in BeamlinesExt.jl test file. Can you add these?
```
We should discuss this. I would rather break up tests into separate files.

mattsignorelli:
@DavidSagan so basically thetre should be a test file called `aperture_test.jl` which has no knowledge of `Beamlines.jl` - instead it should construct a `Coords` and call that with `BeamTracking.launch!` and then check for correctness of the kernel. Also would be good to test for allocations and type stability which can be done easily using `test_matrix` utility function (just make sure the particle coordinates are not lost - this is just because `test_matrix` automatically has these checks. Obiviously the matrix will be the identity).

Then in `test/BeamlinesExt.jl` you can include some tests including a `Beamline`

mattsignorelli:
> We should discuss this. I would rather break up tests into separate files.

This is how it is currently done. Each test file tests a core tracking module, and then `BeamlinesExt` tests the connection of beamlines to that core

src/kernels/aperture.jl
```diff
@@ -0,0 +1,19 @@
...
+
+#
+
+@makekernel fastgtpsa=true function track_aperture_elliptical!(i, coords::Coords, x1, x2, y1, y2)
+  v = coords.v
+
+  x = (2*v[i,XI] - (x2 + x1)) / (x2 - x1)
+  y = (2*v[i,YI] - (y2 + y1)) / (y2 - y1)
+  r = x*x + y*y
+  coords.state[i] = vifelse(r > 1 || x2 < x1 || y2 < y1, STATE_LOST, coords.state[i])
```
JosephPeterDevlin: I don't think this line is going to work with SIMD vectors. In my experience, you are going to need to put parentheses around all the conditions and change `||` to `|`. You should turn `use_KA` to false in `kernel.jl` and try tracking a bunch through an aperture.

DavidSagan:
@JosephPeterDevlin So I changed the code to:
```
  coords.state[i] = vifelse((r > 1) | (x2 < x1) | (y2 < y1), STATE_LOST, coords.state[i])
```
And with SIMD I get the error:
```
Elliptical: Error During Test at /Users/dcs16/.julia/dev/BeamTracking/test/ApertureTracking_test.jl:37
  Got exception outside of a @test
  MethodError: no method matching |(::Vec{2, Bool}, ::Bool)
  The function `|` exists, but no method is defined for this combination of argument types.
  
  Closest candidates are:
    |(::Any, ::Any, ::Any, ::Any...)
     @ Base operators.jl:596
    |(::Bool, ::Bool)
     @ Base bool.jl:39
    |(::Static.False, ::Bool)
     @ Static ~/.julia/packages/Static/SeEGr/src/Static.jl:511
    ...
```  
So one solution is to cast something like `x2 < x1` as a vector instead of a scalar but I don't know the best way to do this. Thoughts?

Note: I checked in my change if you want to test.

JosephPeterDevlin:
@DavidSagan Is it possible to just separate the particle-dependent (`r2`) if-else and the universal (`x1`, etc.) if-else?

DavidSagan:
@JosephPeterDevlin OK that works!

DavidSagan:
@mattsignorelli Ready for another review.

DavidSagan:
@mattsignorelli And `test_matrix` is now used.

mattsignorelli:
Looks good to me! Instead of just doing LOST, would it make sense to do LOST_POS_X, LOST_NEG_X, etc?

DavidSagan:
Yes using `LOST_POS_X`, etc. makes sense. Can we leave this for a later PR?

mattsignorelli (approved):
Approved w tentative work in #161

PR #168, opened 2025-09-14
vifelse for reverse diff support

mattsignorelli:


mattsignorelli:
Note i define Beamtracking's own vifelse to avoid type piracy

PR #169, opened 2025-09-16
Beam-Beam Kernel with test

RevanEoot:
- Added beam-beam kernel
- Added SpecialFunctions and Roots for beam-beam tracking
- Added a single test

The current tests are failing, so I will continue to update until tests pass.


RevanEoot:
Current tests are off by a lot from Bmad. I'm trying to find where the numbers are deviating, but I suspect it's coming from the collision point calculation.

RevanEoot:
Fixefdbeambeam kernel and added extensive tests for particle tracking. The tests don't always pass for phase space z due to a problem with the exact_drift kernel. 
I will soon have tests for the TPS functionality.
I will also be adding the crossing angle.

mattsignorelli:
Can you provide a minimal working example (no beam-beam) where `exact_drift` disagrees with Bmad?

RevanEoot:
Collision point calculation now uses a formula rather than root finder and all tests are passing. TPS tracking agrees on zero orbit, but disagrees off the zero orbit. This is likely a result of something within the kick calculation, but I don't know the exact origin at the moment.

PR #179, opened 2025-10-09
Ramp

JosephPeterDevlin:
@mattsignorelli Let me know if I did anything illegal here. For now, it's fake ramping because ramping with a cavity adds more complications (even in current Bmad).

Also note that I removed the part where changing species was supposed to change the coordinates of the bunch. The normalization is to momentum, not the rigidity, so I don't think this makes sense.


ext/BeamTrackingBeamlinesExt/unpack.jl
```diff
@@ -48,11 +48,23 @@ function universal!(
   # Current KernelChain length is 5 because we have up to
   # 2 aperture, 2 alignment, 1 body kernels
   # TODO: make this 6 when we include update_P0!
-  kc = KernelChain(Val{5}(), RefState(t_ref[], beta_gamma_ref))
+  kc = KernelChain(Val{8}(), RefState(t_ref[], beta_gamma_ref))
```
mattsignorelli: can u update the comment stating why 8 is necessary?

mattsignorelli: nvm i understand. I am making some small changes now including adding t_ref to the Bunch struct itself. Will push soon

PR #192, opened 2025-11-19
Stochastic radiation

JosephPeterDevlin:


JosephPeterDevlin:
@mattsignorelli Any comments?

mattsignorelli:
It looks great... I wonder if we should do a similar thing for radiation damping (make information static). Maybe we should talk about it. Static information is good bc you remove the runtime checks, but also now the compiled binary grows larger which can be a pain, we should discuss the tradeoffs

JosephPeterDevlin:
> It looks great... I wonder if we should do a similar thing for radiation damping (make information static). Maybe we should talk about it. Static information is good bc you remove the runtime checks, but also now the compiled binary grows larger which can be a pain, we should discuss the tradeoffs

It should be straightforward enough to benchmark both. I can do it if you can point me towards a good benchmark

JosephPeterDevlin:
@mattsignorelli I am going to merge this and move discussion of static radiation damping to #194

PR #195, opened 2025-11-23
Ramping improvements

JosephPeterDevlin:


JosephPeterDevlin:
@mattsignorelli I think this is what we decided on for making the fake no-cavity ramping optional

PR #214, opened 2025-12-30
cleanup Float64 usage

mattsignorelli:
This PR changes things like `0.25 * x` to `x / 4` improving performance and also enabling other precision floats for GPU

PR #215, opened 2025-12-31
actually just make F32 when exact

mattsignorelli:


mattsignorelli:
nvm this doesn't work bc e.g. `0.1-0.5*pi != 0.1-0.5f0*pi` so going back to #214

PR #225, opened 2026-01-20
scalar_params option

mattsignorelli:


mattsignorelli:
Uncovered lines are tested in Beamlines so merging

PR #232, opened 2026-01-25
Add macro-particle weight to Coords struct

ndwang:
`weight` is a weighting factor typically used when macro-particles are being simulated. It is optional and defaults to nothing.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -20,16 +20,17 @@ const STATE_LOST_PZ    = UInt8(7)
 const STATE_LOST_Z     = UInt8(8)
 
 # Always SOA
-struct Coords{S,V,Q}
+struct Coords{S,V,Q,W}
   state::S # Array of particle states
   v::V     # Matrix of particle coordinates
   q::Q     # Matrix of particle quaternions if spin else nothing 
-  function Coords(state, v, q)
+  weight::W     # Array of particle weights if weighted else nothing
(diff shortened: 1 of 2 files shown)


mattsignorelli:
In what cases do particles of a bunch (one species) each have different weights?

ndwang:
> In what cases do particles of a bunch (one species) each have different weights?

1. Instead of using macro-particle density to represent a distribution, you can have macro-particles on a fixed grid and use weights. This avoids counting noise.
2. When you have a weak halo, you don't want to track 1 billion particles in the core just so you can have a reasonable number of particles on the edge.

DavidSagan:
Weights can also be used, for example, when simulating space charge, where the weight is used to calculate how large a wake is left by a macroparticle. A typical situation is to have a leading particle with a large weight which generates a reasonable wake followed by "test" particles with zero weights to not perturbe the wake.

mattsignorelli:
sounds good. 

we should probably just use `Base.@kwdef` at this point for coords (requiring only v), and then let the `Bunch` ctor also be all keywords in the same way. This will make initializing a bunch less awkward, e.g. `Bunch(v, nothing, weights)` instead would be `Bunch(v=v, weights=weights)`. However if we also remove the old ctors, we would have to make sure to bump breaking change version

mattsignorelli:
this will be merged once tests all tests pass

PR #234, opened 2026-01-27
Added spin fringe for bend

JosephPeterDevlin:


mattsignorelli:
I texted you this, but I do still have the beating. It turns out it is the solenoid fringe that matters. 

If you think this is good tho, you can merge, bc we do need this anyways

PR #246, opened 2026-02-07
Sagan RFcavity

DavidSagan:
@mattsignorelli @JosephPeterDevlin I still have to add some tests but in the meantime can you take a look and tell me if you see any issues? Thanks.


src/utils/energy.jl
```diff
@@ -1,10 +1,12 @@
 # Particle energy conversions =============================================================
-R_to_E(species::Species, R) = @FastGTPSA sqrt((R*C_LIGHT*chargeof(species))^2 + massof(species)^2)
 E_to_R(species::Species, E) = @FastGTPSA massof(species)*sinh(acosh(E/massof(species)))/C_LIGHT/chargeof(species) 
-pc_to_R(species::Species, pc) = @FastGTPSA pc/C_LIGHT/chargeof(species)
+E_to_c_beta(species::Species, E) = @FastGTPSA sqrt(E^2 - massof(species)^2) * C_LIGHT / E
```
JosephPeterDevlin: Is c_beta not the velocity? The naming should be consistent

src/kernels/multipole.jl
```diff
@@ -15,27 +15,26 @@ multipole magnet. This method supposedly has good numerical
...
 """
 @makekernel fastgtpsa=true function multipole_kick!(i, coords::Coords, ms, knl, ksl, excluding)
-  v = coords.v
-  alive = (coords.state[i] == STATE_ALIVE)
-  bx, by = normalized_field(ms, knl, ksl, v[i,XI], v[i,YI], excluding)
-  bx_0 = zero(bx)
-  by_0 = zero(by)
-  v[i,PXI] -= vifelse(alive, by, by_0)                   
-  v[i,PYI] += vifelse(alive, bx, bx_0)
+  if length(ms) != 0
```
JosephPeterDevlin: You're checking for this in the cavity functions now, right?

ext/BeamTrackingBeamlinesExt/yoshida_bl.jl
```diff
@@ -22,12 +22,6 @@
   end
 end
 
-@inline alignment(tm::AbstractYoshida, bunch, alignmentparams, bendparams, L, entering) =
```
JosephPeterDevlin, line 25: What's going on with aperture and alignment in this PR?

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,311 @@
...
+end
+
+#---------------------------------------------------------------------------------------------------
+# The "outside" drift does not have the pondermotive kick that the "inside" drift does.
+
+@makekernel fastgtpsa=true function sagan_cavity_outside_drift!(i, coords::Coords, 
+                                     radiation_damping_on, q, m_order, Kn, Ks, a, mass, P0c, L)
+  beta0 = P0c / sqrt(P0c^2 + mass^2)
+  gamma0 = P0c / (mass * beta0)
+  length(m_order) > 0 ? has_multipoles = true : has_multipoles = false
```
JosephPeterDevlin: Why not `has_multipoles = (length(m_order > 0)`?

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,311 @@
...
+end
+
+#---------------------------------------------------------------------------------------------------
+
+"""
+    to_energy_coords!(i, coords::Coords, mass, P0c)
+
+Convert from `(z, pz)` to `(c(t_ref - t), E/P0c)` coords.
+"""
+@makekernel fastgtpsa=true function to_energy_coords!(i, coords::Coords, mass, P0c)
```
JosephPeterDevlin: This function and its inverse need to be merged with the ones in rfcavity_kick.jl somehow

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,311 @@
...
+  v[i,PZI] = vifelse(alive, pz, v[i,PZI])
+end
+
+#---------------------------------------------------------------------------------------------------
+"""
+    reference_energy_shift!(i, coords::Coords, P0c_old, dP0c)
+
+Shift coordinates due to a change in reference energy `dE`.
+"""
+@makekernel fastgtpsa=true function reference_energy_shift!(i, coords::Coords, P0c_old, dP0c)
```
JosephPeterDevlin: This needs to be merged with the function in ramp_P0.jl somehow

src/kernels/spin.jl
```diff
@@ -7,6 +7,7 @@ function omega_multipole(i, coords::Coords, a, g, tilde_m, mm, kn, ks, L)
   @FastGTPSA begin @inbounds begin
     v = coords.v
 
+    # Vector potential is (ax, ay, 0)
```
JosephPeterDevlin: The longitudinal vector potential is not zero, it's just not needed in the same way the transverse components are (for getting kinetic momenta)

src/tracking_methods.jl
```diff
@@ -42,12 +43,26 @@ end
...
+#---------------------------------------------------------------------------------------------------
+
+struct SaganCavity
+  n_cell::Int                # Negative => Use approx half wavelength between cells, Zero => single kick.
+  L_active::Float64          # Negative => Use L as the active length.
+  radiation_damping_on::Bool
+  radiation_fluctuations_on::Bool
+
+  function SaganCavity(; n_cell::Int = -1, L_active::Float64 = -1.0, 
+                             radiation_damping_on::Bool=false, radiation_fluctuations_on::Bool=false)
```
JosephPeterDevlin, line 65: Are there radiation fluctuations in the Sagan cavity?

ext/BeamTrackingBeamlinesExt/utils.jl
```diff
@@ -165,7 +174,23 @@ end
 
 #---------------------------------------------------------------------------------------------------
 
-function rf_phi0(rfparams)
+function rf_phi0_calc(rfparams, species)
+  chargeof(species) > 0 ? dphi = 0 : dphi = pi
+
+  if rfparams.zero_phase == PhaseReference.BelowTransition
+    return rfparams.phi0 + 0.5*pi + dphi
```
JosephPeterDevlin: It's better to use pi/2 for different float types, I think

ext/BeamTrackingBeamlinesExt/utils.jl
```diff
@@ -193,4 +218,61 @@ function fringe_out(f::Fringe.T)
...
+  dE_ref = beamlineP.dE_ref
+  E0_ref = E1_ref - dE_ref
+  dt_ref = L_outer/E_to_c_beta(species, E0_ref) + L_outer/E_to_c_beta(species, E1_ref)
+
+  if n_cell == 0
+    L_inner = L_active / 2
+    dt_ref += L_inner/E_to_c_beta(species, E0_ref) + L_inner/E_to_c_beta(species, E1_ref)
+  else
+    for i_step = 1:n_cell
+      E_now_ref = E0_ref + (i_step-0.5) * dE_ref / n_cell
```
JosephPeterDevlin: 1/2, not 0.5

src/kernels/kernel_utils.jl
```diff
@@ -0,0 +1,9 @@
+@makekernel fastgtpsa=true function multipole_and_spin_kick!(i, coords, mm, knL, ksL, a, tilde_m, L)
+  if isnothing(coords.q) || L == 0
```
JosephPeterDevlin: Why is there no spin kick for zero length?

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,311 @@
...
+
+  to_energy_coords!(i, coords, mass, P0c)
+
+  t = t_phi0 + t_ref - v[i,ZI] / C_LIGHT
+  dE = q_voltage * cos(rf_omega * t)
+
+  rad = dE*dE + 2*sqrt(Pc*Pc + mass^2) * dE + Pc*Pc
+  coords.state[i] = vifelse(rad < 0, STATE_LOST_PZ, coords.state[i])
+  alive = (coords.state[i] == STATE_ALIVE)
+  sqrt_rad = vifelse(alive, sqrt(abs(rad)), 1.0)
```
JosephPeterDevlin: Should probably use something like `one(rad)`

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,311 @@
...
+  t = t_phi0 + t_ref - v[i,ZI] / C_LIGHT
+  phase = rf_omega * t
+  ez_field = q_gradient * cos(phase) 
+  dez_dz_field = q_gradient * sin(phase) * rf_omega / C_LIGHT
+  dE = -edge * dez_dz_field * (v[i,XI]*v[i,XI] + v[i,YI]*v[i,YI]) / 4
+
+  rad = dE*dE + 2*sqrt(Pc*Pc + mass^2) * dE + Pc*Pc
+  coords.state[i] = vifelse(rad < 0, STATE_LOST_PZ, coords.state[i])
+  alive = (coords.state[i] == STATE_ALIVE)
+  sqrt_rad = vifelse(alive, sqrt(abs(rad)), 1.0)
```
JosephPeterDevlin: Same comment as before

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,311 @@
...
+  v = coords.v
+  alive = (coords.state[i] == STATE_ALIVE)
+
+  pz = v[i,PZI]
+  Pc = (1 + pz) * P0c
+  
+  to_energy_coords!(i, coords, mass, P0c)
+  t = t_phi0 + t_ref - v[i,ZI] / C_LIGHT
+  phase = rf_omega * t
+  ez_field = q_gradient * cos(phase)
```
JosephPeterDevlin: For sine and cosine together it's more efficient to use `sincos`

src/utils/energy.jl
DavidSagan: Will rename `E_to_v`

ext/BeamTrackingBeamlinesExt/yoshida_bl.jl
DavidSagan: The code is now cleaner since alignment is now defined to be 
```
 alignment(tm, bunch, alignmentparams, bendparams, L, entering)
```
which works for all tracking methods (which is what is wanted 99.99% of the time).

src/kernels/multipole.jl
DavidSagan: Yes but I thought in general it was a good idea to check at this level. If you want I can take this out.

src/kernels/sagan_cavity.jl
DavidSagan: Makes no difference but I will switch to this.

src/kernels/sagan_cavity.jl
DavidSagan: This will involve some work since the arguments are not the same. I will save this for a later PR.

src/kernels/sagan_cavity.jl
DavidSagan: I am merging using my code. The only diff is that if the particle is not alive, the momentums are not shifted.

src/kernels/spin.jl
DavidSagan: Corrected in update.

src/tracking_methods.jl
DavidSagan: Will be in update shortly.

ext/BeamTrackingBeamlinesExt/utils.jl
DavidSagan: Changed in next update.

ext/BeamTrackingBeamlinesExt/utils.jl
DavidSagan: Changed in next update.

src/kernels/kernel_utils.jl
DavidSagan: Mistake that was left in. Corrected. In the early code zero length with finite solenoid strength lead to divide by zero problems.

src/kernels/sagan_cavity.jl
DavidSagan: changed in next revision.

src/kernels/sagan_cavity.jl
DavidSagan: Changed in next revision.

src/kernels/sagan_cavity.jl
DavidSagan: Changed in next revision.

DavidSagan:
@mattsignorelli @JosephPeterDevlin Incorporated Joe's comments so the code is ready for further review. I am still working on code coverage with the tests.

ext/BeamTrackingBeamlinesExt/BeamTrackingBeamlinesExt.jl
```diff
@@ -1,9 +1,12 @@
 module BeamTrackingBeamlinesExt
 using Beamlines, BeamTracking, GTPSA, StaticArrays, KernelAbstractions, AtomicAndPhysicalConstants
 using Beamlines: isactive, deval, unsafe_getparams, isnullspecies
-using BeamTracking: get_N_particle, R_to_beta_gamma, R_to_gamma, R_to_pc, R_to_v, beta_gamma_to_v,
+using BeamTracking: R_to_E, R_to_beta_gamma, R_to_gamma, R_to_pc, R_to_v, 
+                    beta_gamma_to_v, E_to_R, E_to_v,
                     @makekernel, Coords, KernelCall, KernelChain, push, TimeDependentParam, RefState, 
-                    launch!, AbstractYoshida, rot_quaternion, inv_rot_quaternion, atan2
+                    launch!, AbstractYoshida, rot_quaternion, inv_rot_quaternion, atan2, SaganCavity,
```
mattsignorelli:
```suggestion
                    launch!, AbstractYoshida, rot_quaternion, inv_rot_quaternion, atan2, 
```
No need to import something that should be exported

ext/BeamTrackingBeamlinesExt/BeamTrackingBeamlinesExt.jl
```diff
@@ -39,9 +42,12 @@ function track!(
   return bunch
 end
 
+include("rfcavity_bl.jl")
 include("unpack.jl")
 include("scibmadstandard.jl")
 include("exact.jl")
 include("yoshida.jl")
+include("sagan_cavity_bl.jl")
+include("general_bl.jl")
```
mattsignorelli: Why are there `_bl` here?

ext/BeamTrackingBeamlinesExt/rfcavity_bl.jl
mattsignorelli, line 1: Why is this in its own file?

ext/BeamTrackingBeamlinesExt/rfcavity_bl.jl
```diff
@@ -0,0 +1,13 @@
+#---------------------------------------------------------------------------------------------------
+
+#@inline 
+function RFcavity(tm, bunch, bmultipoleparams, rfparams, beamlineparams, L)
+    rf_omega = rf_omega_calc(rfparams, beamlineparams.beamline.line[end].s_downstream, bunch.species, bunch.p_over_q_ref)
+    t_phi0 = rf_phi0_calc_old(rfparams, beamlineparams.beamline.species_ref) / rf_omega
```
mattsignorelli:
```suggestion
```

No calculations should be done until the tracking method is determined. This functionality needs to be moved into the lowest level "unit" functions implemented by each tracking method. If `beamlineparams` is needed, then add it as an argument to the unit function

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
mattsignorelli, line 1: By unit function, i mean you should implement `pure_rf` and `bmultipole_rf`

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
```diff
@@ -0,0 +1,46 @@
+#---------------------------------------------------------------------------------------------------
+
+#@inline 
+function RFcavity(tm::SaganCavity, bunch, bmultipoleP, rfP, beamlineP, L)
```
mattsignorelli:
```suggestion
function RFcavity(tm::SaganCavity, bunch, bmultipoleparams, rfparams, beamlineparams, L)
```

We should have a uniform convention

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
```diff
@@ -0,0 +1,46 @@
...
+  species = bunch.species
+  p1_over_q_ref = beamlineP.beamline.p_over_q_ref
```
mattsignorelli, line 8:
```suggestion
  p1_over_q_ref = beamlineparams.beamline.p_over_q_ref
```

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
```diff
@@ -0,0 +1,46 @@
...
+  rf_omega = rf_omega_calc(rfP, beamlineP.beamline.line[end].s_downstream, species, p1_over_q_ref)
```
mattsignorelli, line 9:
```suggestion
  rf_omega = rf_omega_calc(rfP, beamlineparams.beamline.line[end].s_downstream, species, p1_over_q_ref)
```

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
```diff
@@ -0,0 +1,46 @@
...
+  E1_ref = R_to_E(species, p1_over_q_ref)
+  dE_ref = beamlineP.dE_ref
```
mattsignorelli, line 11:
```suggestion
  dE_ref = beamlineparams.dE_ref
```

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
```diff
@@ -0,0 +1,46 @@
...
+  E0_ref = E1_ref - dE_ref
+  p0_over_q_ref = E_to_R(species, E0_ref)
+  t_phi0 = rf_phi0_calc(rfP, beamlineP.beamline.species_ref) / rf_omega
```
mattsignorelli, line 15:
```suggestion
  t_phi0 = rf_phi0_calc(rfP, beamlineparams.beamline.species_ref) / rf_omega
```

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
```diff
@@ -0,0 +1,46 @@
...
+
+  mass = massof(species)
+  q = chargeof(species)
+  n_cell, L_active = rf_step_calc(tm.n_cell, tm.L_active, rf_omega, L)
+  t_ref = 0    # bunch.t_ref ## Relative time tracking assumed for now.
+  a = gyromagnetic_anomaly(species)
+
+  if L_active == 0
+    if isnothing(bmultipoleP)
```
mattsignorelli, line 39:
```suggestion
    if isactive(bmultipoleparams)
```

`isnothing` is different from `isactive`, this is the benefit of implementing the unit functions as I say above

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
```diff
@@ -0,0 +1,46 @@
...
+      return KernelCall(BeamTracking.sagan_cavity_thin!,
+                   (tm.radiation_damping_on, tm.radiation_fluctuations_on, mass, q, E0_ref, dE_ref,
```
mattsignorelli: You should store `radiation_damping_on` and `radiation_fluctuations_on` in a `Val` type, storing the information at compile time instead. Currently, this makes every single particle check if this is on, which is wasteful

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
```diff
@@ -0,0 +1,46 @@
...
+                   t_ref, emptySA, emptySA, emptySA, a, q*rfP.voltage, rf_omega, t_phi0, L))
+    else
+      BnL, BsL = get_integrated_strengths(bmultipoleP, L, p0_over_q_ref)
+      return KernelCall(BeamTracking.sagan_cavity_thin!, 
+                   (tm.radiation_damping_on,  tm.radiation_fluctuations_on, mass, q, E0_ref, dE_ref,
+                   t_ref, bmultipoleP.order, BnL, BsL, a, q*rfP.voltage, rf_omega, t_phi0, L))
+    end
+
+  else
+    if isnothing(bmultipoleP)
```
mattsignorelli, line 54:
```suggestion
    if isactive(bmultipoleparams)
```

ext/BeamTrackingBeamlinesExt/general_bl.jl
mattsignorelli, line 1: We should not have these in a separate file, like patches they are exact transforms. Instead what we should do is make it so by default all tracking methods use this transform. This is done in my suggestion below in scibmadstandard.jl

ext/BeamTrackingBeamlinesExt/scibmadstandard_bl.jl
```diff
@@ -1,5 +1,3 @@
-@inline alignment(tm::SciBmadStandard, bunch, alignmentparams, bendparams, L, entering) = alignment(Exact(), bunch, alignmentparams, bendparams, L, entering)
-@inline aperture(tm::SciBmadStandard, bunch, apertureparams, entering) = aperture(Exact(), bunch, apertureparams, entering)
 @inline drift(tm::SciBmadStandard, bunch, L) = drift(Exact(), bunch, L)
```
mattsignorelli, line 1:
```suggestion
@inline alignment(tm, bunch, alignmentparams, bendparams, L, entering) = alignment(Exact(), bunch, alignmentparams, bendparams, L, entering)
@inline aperture(tm, bunch, apertureparams, entering) = aperture(Exact(), bunch, apertureparams, entering)
@inline drift(tm::SciBmadStandard, bunch, L) = drift(Exact(), bunch, L) 
```

This makes the DEFAULT for all tracking methods go to the exact transform

mattsignorelli: This should be moved into unpack.jl

ext/BeamTrackingBeamlinesExt/unpack.jl
```diff
@@ -321,7 +316,7 @@ end
 
 
 # === Elements thin vs thick check === #
-@inline pure_rf(tm, bunch, rfparams, omega, t0, L)                          = L == 0 ? thin_pure_rf(tm, bunch, rfparams, omega, t0)                         : thick_pure_rf(tm, bunch, rfparams, omega, t0, L)
+@inline pure_rf(tm, bunch, rfparams, rf_omega, t_phi0, L)                          = L == 0 ? thin_pure_rf(tm, bunch, rfparams, rf_omega, t_phi0)                         : thick_pure_rf(tm, bunch, rfparams, rf_omega, t_phi0, L)
```
mattsignorelli: `t_phi0` ? time-phase 0?

ext/BeamTrackingBeamlinesExt/utils.jl
```diff
@@ -60,12 +62,19 @@ function check_p_over_q_ref!(bl::Beamline, ref, bunch::Bunch, notify=true)
   return
 end
 
+#---------------------------------------------------------------------------------------------------
+
 get_n_multipoles(::BMultipoleParams{T,N}) where {T,N} = N
 
 make_static(a::StaticArray) = SVector(a)
 make_static(a) = a
+const emptySA = SVector{1, Int32}[]
```
mattsignorelli: What is the point of this? Also this is not correct, see:
```julia
julia> typeof(SVector{1,Int32}[])
Vector{SVector{1, Int32}} (alias for Array{SArray{Tuple{1}, Int32, 1, 1}, 1})
```
If you want this just do `SA[]`. But you should not want this

```suggestion
```

ext/BeamTrackingBeamlinesExt/utils_bl.jl
```diff
@@ -193,4 +218,61 @@ function fringe_out(f::Fringe.T)
...
+    rf_step_calc(n_cell, L_active, rf_omega, L) -> n_cell_out, L_active_out
+
+For an element of length `L`, calculate the number of RF cells (kicks) `n_cell_out` and the active length 
+`L_active_out` given the input number of cells `n_cell` and the active length length `L_active`.
+
+If `L_active` is negative, `L_active_out` is set to the element length `L`.
+If `n_cell` is negative, `n_cell_out` is set so that the cell length is near half a wavelength.
+If `L_active` is zero, `n_cell_out` is set to zero.
+"""
+function rf_step_calc(n_cell, L_active, rf_omega, L)
```
mattsignorelli: Note that this will not be SIMD-able over the RF frequency and length, so if someone wants to do `BatchParams` with this cavity model they will not be able to

ext/BeamTrackingBeamlinesExt/utils_bl.jl
```diff
@@ -193,4 +218,61 @@ function fringe_out(f::Fringe.T)
...
+  elseif L_active == 0
+    return 0, L_active
+  else
+    return n_cell, L_act
+  end
+end
+
+#---------------------------------------------------------------------------------------------------
+
+function bunch_dt_ref(tm, bunch, rfP, beamlineP, L)
```
mattsignorelli: Why have this function? It doesn't use `rfparams` and `beamlineparams`?

ext/BeamTrackingBeamlinesExt/yoshida.jl
```diff
@@ -22,12 +22,6 @@
   end
 end
 
-@inline alignment(tm::AbstractYoshida, bunch, alignmentparams, bendparams, L, entering) =
-  alignment(Exact(), bunch, alignmentparams, bendparams, L, entering)
-
-@inline aperture(tm::AbstractYoshida, bunch, apertureparams, entering) =
-  aperture(Exact(), bunch, apertureparams, entering)
-
```
mattsignorelli, line 30: See my comments above

src/kernels/kernel_utils.jl
```diff
@@ -0,0 +1,9 @@
+@makekernel fastgtpsa=true function multipole_and_spin_kick!(i, coords, mm, knL, ksL, a, tilde_m, L)
```
mattsignorelli:
```suggestion
@makekernel function multipole_and_spin_kick!(i, coords, mm, knL, ksL, a, tilde_m, L)
```

src/kernels/kernel_utils.jl
mattsignorelli, line 1: `kernel_utils` is not the best file name for this fundamental function in my opinion. @JosephPeterDevlin input?

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,338 @@
...
+
+@makekernel fastgtpsa=true function sagan_cavity_inside_drift!(i, coords::Coords, 
+                        radiation_damping_on, radiation_fluctuations_on, traveling_wave, 
+                        q, q_gradient, m_order, Kn, Ks, a, mass, P0c, L)
+
+  beta0 = P0c / sqrt(P0c^2 + mass^2)
+  gamma0 = P0c / (mass * beta0)
+
+  # 1/2 pondermotive kick
+  if !traveling_wave
```
mattsignorelli, line 132: This information would be best stored statically (just like I suggest with the radiation) because inside the kernel every single particle has to do this check now

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,338 @@
...
+
+@makekernel fastgtpsa=true function sagan_cavity_thick!(i, coords::Coords, 
+              radiation_damping_on, radiation_fluctuations_on, traveling_wave, 
+              mass, q, E0_ref, dE_ref, t_ref, n_cell, m_order, Bn, Bs, 
+              a, q_voltage, rf_omega, t_phi0, L_active, L)
+
+  L_out = (L - L_active) / 2           # Length outside of active region
+  q_gradient = q_voltage / L_active    # Effective gradient
+  P0c = sqrt(E0_ref^2 - mass^2)
+  q_over_p_ref = q * C_LIGHT / P0c
```
mattsignorelli: These things are the same for all particles? if so they should be computed outside of the kernel in the unpacking step. Kernels should have the maximally processed input parameters so that operations in a kernel are all per-particle

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,338 @@
...
+#---------------------------------------------------------------------------------------------------
+# The "outside" drift does not have the pondermotive kick that the "inside" drift does.
+
+@makekernel fastgtpsa=true function sagan_cavity_outside_drift!(i, coords::Coords, 
+                                     radiation_damping_on, radiation_fluctuations_on, 
+                                     q, m_order, Kn, Ks, a, mass, P0c, L)
+  beta0 = P0c / sqrt(P0c^2 + mass^2)
+  gamma0 = P0c / (mass * beta0)
+  has_multipoles = (length(m_order) > 0)
+  has_sol = (has_multipoles && m_order[1] == 0)
```
mattsignorelli: `has_multipoles` will be done statically because `m_order` is a static array, but `has_sol` will be done for every particle at runtime because `m_order[1] == 0` cannot be done statically. This will hurt performance

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,338 @@
...
+end
+
+#---------------------------------------------------------------------------------------------------
+# The "outside" drift does not have the pondermotive kick that the "inside" drift does.
+
+@makekernel fastgtpsa=true function sagan_cavity_outside_drift!(i, coords::Coords, 
+                                     radiation_damping_on, radiation_fluctuations_on, 
+                                     q, m_order, Kn, Ks, a, mass, P0c, L)
+  beta0 = P0c / sqrt(P0c^2 + mass^2)
+  gamma0 = P0c / (mass * beta0)
```
mattsignorelli, line 155: If these are the same for all particles, compute outside the kernel to not waste time

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,338 @@
...
+@makekernel fastgtpsa=true function to_energy_coords!(i, coords::Coords, beta)
+  v = coords.v
+  alive = (coords.state[i] == STATE_ALIVE)
+  v[i,ZI]  = vifelse(alive, v[i,ZI]/ beta, v[i,ZI])
+  v[i,PZI] = vifelse(alive, (1 + v[i,PZI]) / beta, v[i,PZI])
+end
+
+#---------------------------------------------------------------------------------------------------
+
+"""
```
mattsignorelli, line 332: Hasn't Joe already implemented this functionality? We should reuse code where possible

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,338 @@
...
+@makekernel fastgtpsa=true function to_momentum_coords!(i, coords::Coords, mass, P0c, pz)
+  v = coords.v
+  alive = (coords.state[i] == STATE_ALIVE)
+
+  m_over_pc = mass / ((1 + pz) * P0c)
+  beta = 1 / sqrt(1 + m_over_pc * m_over_pc)
+  v[i,ZI]  = vifelse(alive, beta * v[i,ZI], v[i,ZI])
+  v[i,PZI] = vifelse(alive, pz, v[i,PZI])
+end
```
mattsignorelli, line 346: @JosephPeterDevlin haven't you implemented this already?

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,338 @@
...
+
+#---------------------------------------------------------------------------------------------------
+
+function dpc_given_dE(old_pc, dE, mass)
+  rad = dE*dE + 2*sqrt(old_pc*old_pc + mass^2) * dE + old_pc*old_pc
+  if rad < 0; error("Change in reference energy too negative."); end
+  return sqrt(rad) - old_pc
+end
```
mattsignorelli, line 353: This is not SIMD-able in rad, try `vifelse`

src/kernels/spin.jl
```diff
@@ -77,9 +100,15 @@ function omega_field(i, coords::Coords, a, g, tilde_m, ax, ay, e_vec, b_vec, L)
...
-    oz = (b_perp_z + b_para_z + e_part_z) * L
+    if L != 0
+      ox = (b_perp_x + b_para_x + e_part_x) * L        
+      oy = (b_perp_y + b_para_y + e_part_y + g) * L
+      oz = (b_perp_z + b_para_z + e_part_z) * L
+    else
+      ox = (b_perp_x + b_para_x + e_part_x)
+      oy = (b_perp_y + b_para_y + e_part_y + g)
+      oz = (b_perp_z + b_para_z + e_part_z)
+    end
```
mattsignorelli:
```suggestion
    ox = (b_perp_x + b_para_x + e_part_x) * L        
    oy = (b_perp_y + b_para_y + e_part_y + g) * L
    oz = (b_perp_z + b_para_z + e_part_z) * L
```

we can't ahve these conditionals over the parameters, this is necessary for SIMD and in particular the BatchParams which is one of SciBmad's most important and unique features

src/utils/energy.jl
```diff
@@ -1,10 +1,12 @@
 # Particle energy conversions =============================================================
-R_to_E(species::Species, R) = @FastGTPSA sqrt((R*C_LIGHT*chargeof(species))^2 + massof(species)^2)
 E_to_R(species::Species, E) = @FastGTPSA massof(species)*sinh(acosh(E/massof(species)))/C_LIGHT/chargeof(species) 
-pc_to_R(species::Species, pc) = @FastGTPSA pc/C_LIGHT/chargeof(species)
+E_to_v(species::Species, E) = @FastGTPSA sqrt(E^2 - massof(species)^2) * C_LIGHT / E
```
mattsignorelli, line 3: Why remove `pc_to_R`? 
```suggestion
pc_to_R(species::Species, pc) = @FastGTPSA pc/C_LIGHT/chargeof(species)
E_to_v(species::Species, E) = @FastGTPSA sqrt(E^2 - massof(species)^2) * C_LIGHT / E
```

mattsignorelli (changes requested):

ext/BeamTrackingBeamlinesExt/BeamTrackingBeamlinesExt.jl
DavidSagan: To give a unique name. My thought is to give the other files a similar suffix. We can discuss further.

ext/BeamTrackingBeamlinesExt/rfcavity_bl.jl
DavidSagan: I removed this file and put the code in generl_bl.jl

ext/BeamTrackingBeamlinesExt/rfcavity_bl.jl
DavidSagan: Done in next commit.

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
DavidSagan: Let us talk about uniform naming conventions and implement in a future PR since this making things uniform will touch a fair amount of code.

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
DavidSagan: See above.

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
DavidSagan: See above.

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
DavidSagan: See above.

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
DavidSagan: See above.

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
DavidSagan: See above.

DavidSagan: See above.

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
DavidSagan: See above.

src/kernels/sagan_cavity.jl
DavidSagan: Issue is opened.

src/kernels/sagan_cavity.jl
DavidSagan: This is not a vector operation.

src/utils/energy.jl
DavidSagan: Grouping has changed this.

ext/BeamTrackingBeamlinesExt/sagan_cavity_bl.jl
DavidSagan: Done.

ext/BeamTrackingBeamlinesExt/unpack.jl
DavidSagan: The time that `phi0` represents.

ext/BeamTrackingBeamlinesExt/utils.jl
DavidSagan: @mattsignorelli If I use `SA[]` I get the following error:
```
julia> include("test/runtests.jl")
Precompiling BeamTrackingBeamlinesExt...
  1 dependency successfully precompiled in 2 seconds. 85 already precompiled.
BeamTracking.REGISTER_SIZE = static(16)
Sys.ARCH = :aarch64
SaganCavity: Error During Test at /Users/dcs16/.julia/dev/BeamTracking/test/BeamlinesExt/beamlines_sagan_cavity_test.jl:8
  Got exception outside of a @test
  Tuple field type cannot be Union{}
  Stacktrace:
    [1] _broadcast
      @ ~/.julia/packages/StaticArrays/IZgIP/src/broadcast.jl:115 [inlined]
    [2] copy
      @ ~/.julia/packages/StaticArrays/IZgIP/src/broadcast.jl:60 [inlined]
    [3] materialize
      @ ./broadcast.jl:872 [inlined]
    [4] macro expansion
      @ ~/.julia/dev/BeamTracking/src/kernels/sagan_cavity.jl:55 [inlined]
    [5] sagan_cavity_thick!
      @ ~/.julia/dev/BeamTracking/src/kernel.jl:223 [inlined]
    [6] macro expansion
      @ ~/.julia/packages/Unrolled/KtWta/src/Unrolled.jl:47 [inlined]
    [7] macro expansion
      @ ~/.julia/dev/BeamTracking/src/kernel.jl:63 [inlined]
    [8] macro expansion
      @ ~/.julia/packages/Unrolled/KtWta/src/Unrolled.jl:127 [inlined]
    [9] __generic_kernel!(i::Int64, coords::Coords{Vector{UInt8}, Matrix{Float64}, Nothing}, chain::Tuple{KernelCall{typeof(BeamTracking.sagan_cavity_thick!), Tuple{Val{false}, Val{false}, Val{false}, Float64, Float64, Float64, Float64, Int64, Int64, SVector{0, Union{}}, SVector{0, Union{}}, SVector{0, Union{}}, Vararg{Float64, 6}}}, Vararg{KernelCall{typeof(BeamTracking.blank_kernel!), Tuple{}}, 5}}, ref::BeamTracking.RefState{Float64, Float64})
      @ BeamTracking ~/.julia/packages/Unrolled/KtWta/src/Unrolled.jl:127
   [10] _generic_kernel!(i::Int64, coords::Coords{Vector{UInt8}, Matrix{Float64}, Nothing}, kc::BeamTracking.KernelChain{Tuple{KernelCall{typeof(BeamTracking.sagan_cavity_thick!), Tuple{Val{false}, Val{false}, Val{false}, Float64, Float64, Float64, Float64, Int64, Int64, SVector{0, Union{}}, SVector{0, Union{}}, SVector{0, Union{}}, Vararg{Float64, 6}}}, Vararg{KernelCall{typeof(BeamTracking.blank_kernel!), Tuple{}}, 5}}, BeamTracking.RefState{Float64, Float64}})
      @ BeamTracking ~/.julia/dev/BeamTracking/src/kernel.jl:60
   [11] #launch!#320
      @ ~/.julia/dev/BeamTracking/src/kernel.jl:131 [inlined]
   [12] launch!(coords::Coords{Vector{UInt8}, Matrix{Float64}, Nothing}, kc::BeamTracking.KernelChain{Tuple{KernelCall{typeof(BeamTracking.sagan_cavity_thick!), Tuple{Val{false}, Val{false}, Val{false}, Float64, Float64, Float64, Float64, Int64, Int64, SVector{0, Union{}}, SVector{0, Union{}}, SVector{0, Union{}}, Vararg{Float64, 6}}}, Vararg{KernelCall{typeof(BeamTracking.blank_kernel!), Tuple{}}, 5}}, BeamTracking.RefState{Float64, Float64}})
      @ BeamTracking ~/.julia/dev/BeamTracking/src/kernel.jl:92
   [13] universal!(coords::Coords{Vector{UInt8}, Matrix{Float64}, Nothing}, tm::SaganCavity, ramp_without_rf::Bool, bunch::Bunch{Float64, Float64, Coords{Vector{UInt8}, Matrix{Float64}, Nothing}}, L::Float64, p_over_q_ref::Float64, alignmentparams::Nothing, bendparams::Nothing, bmultipoleparams::Nothing, patchparams::Nothing, apertureparams::Nothing, rfparams::RFParams{Float64}, beamlineparams::BeamlineParams, mapparams::Nothing; kwargs::@Kwargs{})
      @ BeamTrackingBeamlinesExt ~/.julia/dev/BeamTracking/ext/BeamTrackingBeamlinesExt/unpack.jl:242
   [14] universal!(coords::Coords{Vector{UInt8}, Matrix{Float64}, Nothing}, tm::SaganCavity, ramp_without_rf::Bool, bunch::Bunch{Float64, Float64, Coords{Vector{UInt8}, Matrix{Float64}, Nothing}}, L::Float64, p_over_q_ref::Float64, alignmentparams::Nothing, bendparams::Nothing, bmultipoleparams::Nothing, patchparams::Nothing, apertureparams::Nothing, rfparams::RFParams{Float64}, beamlineparams::BeamlineParams, mapparams::Nothing)
      @ BeamTrackingBeamlinesExt ~/.julia/dev/BeamTracking/ext/BeamTrackingBeamlinesExt/unpack.jl:44
   [15] _track!(coords::Coords{Vector{UInt8}, Matrix{Float64}, Nothing}, bunch::Bunch{Float64, Float64, Coords{Vector{UInt8}, Matrix{Float64}, Nothing}}, ele::LineElement, tm::SaganCavity, scalar_params::Bool, ramp_without_rf::Bool; kwargs::@Kwargs{})
      @ BeamTrackingBeamlinesExt ~/.julia/dev/BeamTracking/ext/BeamTrackingBeamlinesExt/unpack.jl:40
   [16] _track!(coords::Coords{Vector{UInt8}, Matrix{Float64}, Nothing}, bunch::Bunch{Float64, Float64, Coords{Vector{UInt8}, Matrix{Float64}, Nothing}}, ele::LineElement, tm::SaganCavity, scalar_params::Bool, ramp_without_rf::Bool)
      @ BeamTrackingBeamlinesExt ~/.julia/dev/BeamTracking/ext/BeamTrackingBeamlinesExt/unpack.jl:2
   [17] track!(bunch::Bunch{Float64, Float64, Coords{Vector{UInt8}, Matrix{Float64}, Nothing}}, ele::LineElement; scalar_params::Bool, ramp_particle_energy_without_rf::Bool, kwargs::@Kwargs{})
      @ BeamTrackingBeamlinesExt ~/.julia/dev/BeamTracking/ext/BeamTrackingBeamlinesExt/BeamTrackingBeamlinesExt.jl:22
   [18] track!(bunch::Bunch{Float64, Float64, Coords{Vector{UInt8}, Matrix{Float64}, Nothing}}, ele::LineElement)
      @ BeamTrackingBeamlinesExt ~/.julia/dev/BeamTracking/ext/BeamTrackingBeamlinesExt/BeamTrackingBeamlinesExt.jl:14
   [19] macro expansion
      @ ~/.julia/dev/BeamTracking/test/BeamlinesExt/beamlines_sagan_cavity_test.jl:11 [inlined]
   [20] macro expansion
      @ ~/.julia/juliaup/julia-1.11.7+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/Test/src/Test.jl:1709 [inlined]
   [21] top-level scope
      @ ~/.julia/dev/BeamTracking/test/BeamlinesExt/beamlines_sagan_cavity_test.jl:9
   [22] include(fname::String)
      @ Main ./sysimg.jl:38
   [23] macro expansion
      @ ~/.julia/dev/BeamTracking/test/BeamlinesExt_test.jl:3 [inlined]
   [24] macro expansion
      @ ~/.julia/juliaup/julia-1.11.7+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/Test/src/Test.jl:1709 [inlined]
   [25] top-level scope
      @ ~/.julia/dev/BeamTracking/test/BeamlinesExt_test.jl:3
   [26] include(fname::String)
      @ Main ./sysimg.jl:38
   [27] top-level scope
      @ ~/.julia/dev/BeamTracking/test/runtests.jl:203
   [28] include(fname::String)
      @ Main ./sysimg.jl:38
   [29] top-level scope
      @ REPL[1]:1
```

ext/BeamTrackingBeamlinesExt/utils_bl.jl
DavidSagan: This was done for you so that unpack would not depend explicitly on any tracking method.

src/kernels/kernel_utils.jl
DavidSagan: OK just suggest a better name.

src/kernels/sagan_cavity.jl
DavidSagan: Done.

ext/BeamTrackingBeamlinesExt/scibmadstandard_bl.jl
DavidSagan: Why make things more complicated? This is not needed.

ext/BeamTrackingBeamlinesExt/utils_bl.jl
DavidSagan: Is this really not compatible with `BatchParams` but all the `L == 0 ?` constructs in unpack.jl are?

ext/BeamTrackingBeamlinesExt/yoshida.jl
DavidSagan: Yes.

src/kernels/sagan_cavity.jl
JosephPeterDevlin: With the advent of batch parameters, almost everything is (potentially) a vector operation.

src/kernels/kernel_utils.jl
DavidSagan: Question: When is setting `fastgtpsa = true` appropriate?

src/kernels/kernel_utils.jl
DavidSagan: My idea is to have a file with "low level" functions that are used by other kernel functions.

src/kernels/sagan_cavity.jl
```diff
@@ -0,0 +1,338 @@
...
+  v[i,ZI]  = vifelse(alive, beta * v[i,ZI], v[i,ZI])
+  v[i,PZI] = vifelse(alive, pz, v[i,PZI])
+end
+
+#---------------------------------------------------------------------------------------------------
+
+function dpc_given_dE(old_pc, dE, mass)
+  rad = dE*dE + 2*sqrt(old_pc*old_pc + mass^2) * dE + old_pc*old_pc
+  if rad < 0; error("Change in reference energy too negative."); end
+  return sqrt(rad) - old_pc
```
JosephPeterDevlin: Is this not ill-conditioned for small energy changes?

src/kernels/spin.jl
DavidSagan: This has been fixed.

src/kernels/sagan_cavity.jl
JosephPeterDevlin: One way to fix this is to use the identity `sqrt(old_pc^2 + x^2) - old_pc = x^2/(old_pc + sqrt(old_pc^2 + x^2))`

DavidSagan: Done.

src/kernels/sagan_cavity.jl
DavidSagan: Actually the check is unnecessary and is removed on next push.

src/kernels/sagan_cavity.jl
DavidSagan: fixed in next push.

ext/BeamTrackingBeamlinesExt/unpack_bl.jl
```diff
@@ -37,13 +37,14 @@ function _track!(
   end
 
   # Function barrier
-  universal!(coords, tm, ramp_without_rf, bunch, L, p_over_q_ref, ap, bp, bm, pp, dp, rp, lp, mp; kwargs...)
+  universal!(coords, tm, ele.name, ramp_without_rf, bunch, L, p_over_q_ref, ap, bp, bm, pp, dp, rp, lp, mp; kwargs...)
```
mattsignorelli: Wait this is new, why are you passing the name now?

ext/BeamTrackingBeamlinesExt/unpack_bl.jl
```diff
@@ -120,21 +122,14 @@ function universal!(
       kc = push(kc, @inline(pure_patch(tm, bunch, patchparams, L)))
     end
 
-  elseif isactive(rfparams)
-    !rfparams.is_crabcavity || error("Crab cavities not yet supported for tracking")
-    omega = rf_omega(rfparams, beamlineparams.beamline.line[end].s_downstream, bunch.species, bunch.p_over_q_ref)
-    t0 = rf_phi0(rfparams) / omega
-
+  elseif !isnothing(rfparams)   # Need to handle the case if Voltage = 0 but dE_ref is finite
```
mattsignorelli: We should only use the `isactive` functions in unpack, in case we do want to revive BitsBeamline. if `isactive(rfparams)` is false then one can still handle nonzero dE_ref, it just won't be in this part of the if-else tree

ext/BeamTrackingBeamlinesExt/unpack_bl.jl
```diff
@@ -244,11 +239,12 @@ function universal!(
...
 
-  # Evolve time through whole element
-  bunch.t_ref += L/beta_gamma_to_v(beta_gamma_ref)
-  
   # noinline necessary here for small binaries and faster execution
   @noinline launch!(coords, kc; kwargs...)
+
+  # Evolve time through whole element
+  bunch.t_ref += bunch_dt_ref(tm, bunch, rfparams, beamlineparams, L)
+
```
mattsignorelli, line 245: Is this also new? I don't see why rfparams is special, I thought we agreed you would add/subtract it in your unit function to account for the default

ext/BeamTrackingBeamlinesExt/utils_bl.jl
mattsignorelli: in your SaganCavity rfcavity unit function, you have the bunch struct, so you can add/subtract whatever you want to the time

src/kernels/kernel_utils.jl
mattsignorelli: @JosephPeterDevlin thoughts?

src/tracking_methods.jl
```diff
@@ -42,12 +43,26 @@ end
...
     return new(fringe_at)
   end
-end
\ No newline at end of file
+end
+
+#---------------------------------------------------------------------------------------------------
+
+struct SaganCavity
+  n_cell::Int                # Negative => Use approx half wavelength between cells, Zero => single kick.
```
mattsignorelli: `n_cells` instead of `n_cell`?

mattsignorelli (changes requested):

src/tracking_methods.jl
JosephPeterDevlin: If we are going for consistency with other tracking methods, it should probably even be `num_cells`

src/kernels/multipole.jl
JosephPeterDevlin: I would take it out if it's no longer necessary.

src/kernels/kernel_utils.jl
JosephPeterDevlin: A good name for a general file is not coming to me at the moment. I think this function would be at home in `multipole.jl`, though.

src/kernels/multipole.jl
DavidSagan: Done in next push.

ext/BeamTrackingBeamlinesExt/unpack_bl.jl
DavidSagan: To use in error messages. I have taken this out for now and will reinstate in a later PR.

src/tracking_methods.jl
DavidSagan: Changing to num_cells in next push.

ext/BeamTrackingBeamlinesExt/scibmadstandard_bl.jl
DavidSagan: Joe will sort this out.

src/kernels/sagan_cavity.jl
DavidSagan: Done.

src/kernels/sagan_cavity.jl
DavidSagan: Done.

src/kernels/kernel_utils.jl
DavidSagan: Resolved.

ext/BeamTrackingBeamlinesExt/utils_bl.jl
DavidSagan: This function has been removed.

src/kernels/kernel_utils.jl
DavidSagan: Can be renamed in a future PR.

ext/BeamTrackingBeamlinesExt/unpack_bl.jl
DavidSagan: Fixed.

ext/BeamTrackingBeamlinesExt/unpack_bl.jl
DavidSagan: Fixed.

mattsignorelli (approved):

src/kernels/spin.jl
```diff
@@ -16,20 +17,60 @@ function omega_multipole(i, coords::Coords, a, g, tilde_m, mm, kn, ks, L)
     end
 
     bx, by = normalized_field(mm, kn, ks, v[i,XI], v[i,YI], -1)
-    bz_0 = zero(kn[1])
-    bz = mm[1] == 0 ? kn[1] : bz_0
-    b_vec = (bx, by, bz)
-    e_vec = (bz_0, bz_0, bz_0)
+    zero_0 = zero(kn[1])
+    if mm[1] == 0 && L != 0
```
JosephPeterDevlin: Why does the length matter here? Also note to self: make this solenoid check static too.

ext/BeamTrackingBeamlinesExt/unpack_bl.jl
```diff
@@ -69,7 +70,8 @@ function universal!(
     p_over_q_ref_initial = bunch.p_over_q_ref
     p_over_q_ref_final = p_over_q_ref(bunch.t_ref)
     if !(p_over_q_ref_initial ≈ p_over_q_ref_final)
-      kc = push(kc, KernelCall(BeamTracking.update_P0!, (p_over_q_ref_initial, p_over_q_ref_final, ramp_without_rf)))
+      kc = push(kc, KernelCall(BeamTracking.reference_momentum_shift!, (p_over_q_ref_initial,
```
JosephPeterDevlin, line 73: It seems like there could be some catastrophic cancellation here. It looks like we only need to work with the ratio initial rigidity over final rigidity, and then there would be no problem.

DavidSagan: @JosephPeterDevlin If you look at the `update_P0!` routine you will see there was also a problem there. This way at least the sagan cavity code (which uses `reference_momentum_shift!`) avoids roundoff.

JosephPeterDevlin: The problem was made worse because now the two values are being subtracted before even entering the kernel.

src/kernels/spin.jl
DavidSagan: Fixed. Sorry this was a leftover from a previous attempt at handling zero length situations.

ext/BeamTrackingBeamlinesExt/unpack_bl.jl
DavidSagan: Why is that worse than having the values subtracted after entering the kernel? I could reinstate `update_P0!` but I want to keep `reference_momentum_shift!` for the cavity tracking since that entirely avoids the problem. What do you want to do?

JosephPeterDevlin: Because before only pz had a problem, but now px and py have a problem too. If you pass the ratio initial rigidity over final rigidity to `reference_momentum_shift!` (instead of initial and difference) then all problems can be avoided.

DavidSagan: I would think that operations before entering the kernel are preferable.

JosephPeterDevlin: For efficiency yes, but I mean that the catastrophic cancellation occurs before the kernel now and it has no chance of getting any of the momenta right, when before the cancellation only occurred pz.

JosephPeterDevlin:
```
@makekernel fastgtpsa=true function reference_momentum_shift!(i, coords::Coords, P0c_ratio, shift_pz)
  v = coords.v
  alive = (coords.state[i] == STATE_ALIVE)

  v[i,PXI] = vifelse(alive, P0c_ratio * v[i,PXI], v[i,PXI])
  v[i,PYI] = vifelse(alive, P0c_ratio * v[i,PYI], v[i,PYI])
  if shift_pz
    v[i,PZI] = vifelse(alive, P0c_ratio * (1 + v[i,PZI]) - 1, v[i,PZI])
  end
end
```
What is the problem with this?

DavidSagan: Problem is roundoff with pz.

DavidSagan: And px and py do not have a problem.

JosephPeterDevlin: The existing code subtracts O(p_over_q_ref) numbers and this subtracts O(1) numbers.

Issue #213, opened 2025-12-30
Selecting between relative and absolute time tracking


DavidSagan:
The switch also needs to select between using absolute time as opposed to the absolute time relative to the first turn reference time (see the Bmad documentation on `bmad_com%absolute_time_ref_shift`).

DavidSagan:
Here is how the three time modes can be implemented given a selector switch. In the followinging, `t_phase` is the time used in the phase calculation:
```
  t_rf = phi0/omega + t_phase - z/(c*beta)
  phase = omega * t_rf
```
Let `t_ref1` be the reference time computed for a given RF element on the first turn and `t_refN` be the (accumulated) reference time computed for the element at turn `N`. At turn `N`, `t_phase` will be:
```
  t_phase = 0                   # Relative time
  t_phase = t_refN - t_ref1     # Absolute time  with ref shift (bmad_com%absolute_time_ref_shift = T which is the default)
  t_phase = t_refN             # Absolute time, no ref shift (bmad_com%absolute_time_ref_shift = F)
```

mattsignorelli:
I have some ideas... but first a question: Is there ever a case where you might want relative time tracking in some elements and absolute in others?

DavidSagan:
With Bmad, no one has ever wanted mixed relative and absolute time tracking.

mattsignorelli:
Correct me if I'm wrong, but wouldn't mixing the two be super useful for studies like resonant depolarization? The RF cavities all around the ring would stay relative time tracking, but the AC kicker would be absolute

mattsignorelli:
I know having to set the phases for the rf cavities around the ring in FCC RDP studies has been a major pain point in Bmad

DavidSagan:
Interesting. We should discuss.

mattsignorelli:
I don't think handling it would be too hard. we can just have a flag in the RFParams, and then in the Time struct one could specify `Time(absolute=false)` for example. This would be super easy to implement. Mixing the two inside a single time function, e.g.
```
Time(absolute=false) + Time()
```
Is not as trivial, but i think we can find a solution. This also I imagine would be rare

DavidSagan:
Thought: To get different types of time tracking for different elements, have a master switch that sets the default and have, for each RF element, a local switch that can override the default.

eiad-hamwi:
Two possible routes: 

1) Some Beamlines parameter group could contain `absolute_time_tracking::Bool`
[...]

PR #250, opened 2026-02-19
Extending batch lowering/evaluation for structs

ndwang:
## Summary

Since the introduction of TimeDependentParam and BatchParam, the kernel launch sequence is no longer transparent and now implicitly changes the kernel arguments. The batch lowering/evaluation logic previously only supports tuples and plain BatchParams. This PR extends that logic to support structs.

- Extend `batch_lower`, `static_batchcheck`, and `beval` to recursively handle structs containing BatchParam fields
- Add passthrough methods for known non-batch types (TimeDependentParam, Funciton, SArray)
- Add tests for struct lowering/evaluation including nested structs

## Limitation
The method does not support types that:
1. A field does contain BatchParam, and
2. The type has parameters that aren't inferable from fields alone

For example, SVector{S, T} where S is the array shape needs an explicit method if they are allowed to contain BatchParam.

Structs with shared type parameters (e.g. `struct S{T}; a::T; b::T; end`) will fail if the two fields lower to different types.

## Tested
- Tuples and plain BatchParam behaviors are unchanged
- Struct with BatchParam field lowers and evaluates correctly
- Nested structs with BatchParam fields work correcly


ndwang:
TimeDependentParam contains a function which is considered a struct and hits limitation 1. And this probably won't work well with GTPSA. 
Implementing a more general method to work with types whose parameters encode non-field info

ndwang:
Now this should work in general unless people are doing BatchParam on TPS or other wild combinations, then it needs to be addressed case by case

src/batch.jl
```diff
@@ -281,15 +281,46 @@ function batch_lower(b::BatchParam)
   end
 end
 
-batch_lower(bp) = bp
```
mattsignorelli, line 284: This was the passthrough, what was wrong with this?

src/batch.jl
```diff
@@ -281,15 +281,46 @@ function batch_lower(b::BatchParam)
...
+# Passthrough methods for types that never contain BatchParam.
+# These short-circuit before entering _batch_lower_struct for performance.
+batch_lower(bp::TimeDependentParam) = bp
+batch_lower(bp::Function) = bp
+batch_lower(bp::SArray{N,T}) where {N,T} = bp
+
+# Recursively lower BatchParam fields in structs.
+batch_lower(bp::T) where {T} = _batch_lower_struct(bp)
+
+function _batch_lower_struct(bp::T) where {T}
```
mattsignorelli: This function works, but it would be more performant to instead use generated functions, similar to what I do here where I apply a function `deval` and `scalarize` to each element of a struct and reconstruct it: https://github.com/bmad-sim/Beamlines.jl/blob/4c753d3563bf38571653d70e2c40b950223caa2b/src/element.jl#L5-L31 

Here it would be the same thing, but with `batch_lower`

Things like line 306 may be slow bc indexing a tuple, and I think `fieldnames` can actually also be slow :/

also unfortunately doing `T.name.wrapper(splat...)` is also slow :(

src/batch.jl
```diff
@@ -299,6 +330,18 @@ static_batchcheck(::_LoweredBatchParam) = true
   end
   return false
 end
+# Recursively check for batch params inside structs.
+function static_batchcheck(s::T) where {T}
```
mattsignorelli: A generated function is the safest bet to ensure the check is static. Here it has to get each field of the struct and check, which may end up being at runtime

src/batch.jl
```diff
@@ -332,17 +375,35 @@ function lane2vec(lane::SIMD.VecRange{N}) where {N}
   end
 end
 
-@inline beval(b, i) = b
```
mattsignorelli, line 335: This was the passthru

mattsignorelli:
While we could try to provide a generic `batch_lower` for structs using generated functions, since all structs (and their (inner) constructors) can be different, I think the safest bet is to just require structs to define their own `batch_lower`  and `beval`

ndwang:
The old catch-all won't act on BatchParam inside a struct, which is the whole point of this PR. The new `_batch_lower_struct` provides a generic method for structs and immediately returns for non-struct types. This handles the most common cases in kernels

If there's no BatchParam, `_batch_lower_struct` skips reconstruction, which is the same behavior as before. In the rare case where someone needs BatchParam inside a struct with an inner constructor with a different signature, `batch_lower` fails with a clear MethodError. This is safer than the old behavior where BatchParam inside a struct would go through untouched and cause failure in the kernel

The launch! sequence was supposed to be transparent and no one needs to look inside. Requiring every struct to define its own `batch_lower` is a dangerous pattern as we are leaking tracking kernels into batch logic. Unpacking, launch, and kernels are coupled together and we no longer have modularity

Using generated functions for `batch_lower` and `static_batchcheck` is a good point. I did that for `beval` but I need to check how to make it work for `batch_lower`

mattsignorelli:
OK yes. So basically you want the catch-all to instead first check if the type is a struct, and if it is attempt to convert the `BatchParam` fields `_LoweredBatchParam` by default. If not, then just pass through. I think that is a nice feature too. We just have to be careful that everything, especially the `static_batchcheck` which is called inside a kernel itself is _actually_ static, since I've found that working with these "static reflection" functions can be deceiving and not quite static.

A best, most careful approach for the `batch_lower` part would be to use `ConstructionBase.jl` which ensures we can reconstruct a given object with all of its fields/properties accordingly `batch_lower`-ed. This also has taken care of a lot of the annoying parts with "static" reflection. I don't think that actually is too hard given its `setproperties` interface etc: https://juliaobjects.github.io/ConstructionBase.jl/dev/#type-tips 

The harder part will be the `static_batchcheck` , ensuring it is actually fully static. With generated functions and `ConstructionBase` I believe it is doable, but it is challenging, hence why we've resorted to just passing parameters as tuples for now since its "easy". But it would be great if you can get a solution for that!

mattsignorelli:
And then, whatever solution you have for this, we can have for `TimeDependentParam` too since its structure is very similar to `BatchParam`

mattsignorelli:
@ndwang this looks good, can you do the same thing for `TimeDependentParam`? Then RK4 will also work for time varying multipoles, which is needed. The structure is very similar to batch

ndwang:
@mattsignorelli I have the same thing ready for `TimeDependentParam`. I'll push it today after I add tests.
I'm wondering how time should behave between particle time, RF time, time dependent parameter, and multi-step tracking. Let's discuss at the Friday meeting

mattsignorelli:
> I'm wondering how time should behave between particle time, RF time, time dependent parameter, and multi-step tracking. Let's discuss at the Friday meeting

At the moment, `Time()` is always the absolute time. That is handled internally. There will soon be a way to specify if you want it to be relative time (see #213) . This won't affect your implementation, though.

RF time is currently always relative time

PR #260, opened 2026-03-09
Bumpbl

mattsignorelli:
@DavidSagan @JosephPeterDevlin I cleaned up the RF stuff a bit to make things consistent with the new Beamlines, but Yoshida does not work with the `rf_phi0_calc`. Can you resolve this?


mattsignorelli:
OK @JosephPeterDevlin I might have fixed your phase factor, I had to subtract pi/2 from the output of rf_phi0_calc to get the matrix in the lattice `test/lattices/esr.jl` to agree. 

However, @DavidSagan , your matrix does not agree with the ESR's matrix. This is unrelated to the current changes made here, so that might be worth investigating

mattsignorelli:
@JosephPeterDevlin  do you have any cavity tests with different species (proton)? To account for the zero_phase difference between particles with + and - charge?

mattsignorelli:
Also @JosephPeterDevlin  it seems like the current tests are failing possibly because the `zero_phase` is now specified consistently, and should now be specified for your lattice elements in the tests

DavidSagan:
> @JosephPeterDevlin do you have any cavity tests with different species (proton)? To account for the zero_phase difference between particles with + and - charge?

Note: With `rf_phi0_calc`, the setting of phi0 should not have to be changed when, for example, changing between e+ and e-.

DavidSagan:
> However, @DavidSagan , your matrix does not agree with the ESR's matrix. This is unrelated to the current changes made here, so that might be worth investigating

How are you setting up the cavity element when you do the changes and how far off is the matrix?

PR #261, opened 2026-03-09
Correct R_to_v function.

DavidSagan:
Since `R = P0/q` need to have a factor `q` in the function for things like `He++`.
I will add a test shortly.


mattsignorelli:
~~I corrected this already in #260~~

I see now, the factor of q still needs to be there

mattsignorelli (approved):

PR #278, opened 2026-03-30
add callback functionality and clean Yoshida

### LINKED ISSUE #256: Emulate MAD-NG's "actions"
In MAD-NG, "actions" are functions that can be applied at each step of the integrator. it is fully generic, and so this is how Laurent gets Twiss parameters inside of elements. This is basiclly the same as the "integrals" idea we had to compute e.g. radiative envelope and DK polarization, but more general in that even the Twiss call uses actions. I think this is a great structure and we should follow suit

mattsignorelli:
Also included is a new `Bunch` constructor that is fully kwarg


ext/BeamTrackingBeamlinesExt/yoshida_bl.jl
```diff
@@ -291,38 +287,43 @@ end
...
 # =========== BENDING ELEMENTS ============= #
-@inline thick_bend_no_field(tm::Union{Yoshida,BendKick}, bunch, bendparams, L) = 
-  thick_bend_no_field(Exact(), bunch, bendparams, L)
+@inline function thick_bend_no_field(tm::Union{Yoshida,BendKick}, bunch, bendparams, L)
+  g = bendparams.g_ref
+  tilt = bendparams.tilt_ref
+  e1 = bendparams.e1
+  e2 = bendparams.e2
+  w = rot_quaternion(0,0,-tilt)
+  w_inv = inv_rot_quaternion(0,0,-tilt)
```
mattsignorelli: @JosephPeterDevlin I noticed this tilt is negative while other bend tilts are positive, is this intentional for the `exact_curved_drift!` kernel?

mattsignorelli: I think this is a problem since it calls the same bend kernel under the hood

mattsignorelli:
PR also updates quaternion utility functions to make tuples for batch angle support

ext/BeamTrackingBeamlinesExt/yoshida_bl.jl
JosephPeterDevlin: I think I see what happened. The negative sign is right here. The only typo was in Yoshida's thick_bend_pure_bdipole, but this was never noticed, because it's only called when spin or radiation is on and the rotated bend test doesn't have either. The other positive tilts are from rotated multipoles, not tilt_ref.

mattsignorelli: So is this PR ready to go or do we have stuff to fix still?

JosephPeterDevlin: Yoshida's thick_bend_pure_bdipole has the wrong sign for the tilt angle

mattsignorelli: `photon_params` also receives the tilt, does that get flipped too?

JosephPeterDevlin: Yes

mattsignorelli: same thing for `thick_bend_bdipole`?

JosephPeterDevlin: Yes

JosephPeterDevlin:
@mattsignorelli Are curved drifts and bends with multipoles still supposed to be treated differently in `compute_g`?

mattsignorelli:
I believe I handled that properly, you can check. nothing matters except `g_ref` and `tilt_ref`, its entirely coordinate system dependent

JosephPeterDevlin:
> I believe I handled that properly, you can check. nothing matters except `g_ref` and `tilt_ref`, its entirely coordinate system dependent

It looks like the two cases in `compute_g` still use opposite signs. I thought it would be the same

mattsignorelli:
Oh i think you're right

PR #303, opened 2026-06-18
Change tracking species to electron in the SaganCavity tests to avoid…

DavidSagan:
… the fact that the gyromagnetic anomaly is not defined for Helium.


mattsignorelli:
How was it working before?

DavidSagan:
I'm guessing that Bmad and SciBmad give the same garbage number for the gyromagnetic anomaly. Either that or I switched from using electrons to #3He after checking with Bmad. In either case, changing to electrons is the right thing to do.

DavidSagan:
Talked to Matt and he approves.

PR #304, opened 2026-06-19
Adding fixed-point iteration option to implicit

JosephPeterDevlin:
In the cases I have tested, the fixed-point iteration is faster, but the advantage depends on how expensive each potential/Jacobian evaluation is. I am leaving it as a flag, but setting the fixed-point iteration to be default.


JosephPeterDevlin:
Need to test on GPU before merging

PR #315, opened 2026-07-17
Add PythonCall extension: TimeDependentParam/BatchParam from Python

DavidSagan:
Part of the Python-interface work for bmad-sim/SciBmad.jl#76.

## What

New `BeamTrackingPythonCallExt` package extension so BeamTracking's deferred-parameter types can be constructed from Python objects handed over via juliacall/PythonCall:

```python
TimeDependentParam(lambda t: 2*t + 1)   # from a Python callable
TimeDependentParam(5.0)                 # constant
BatchParam([0.1, 0.2, 0.3])             # from a Python sequence
```

Adds `PythonCall` as a weakdep + `[extensions]` entry + `[compat]` in `Project.toml`, and the extension module with:

- `TimeDependentParam(::Py[, ::Type{T}])`
- `BatchParam(::Py[, ::Type{T}])`

converting values to `T` (default `Float64`).

## Why

`TimeDependentParam` and `BatchParam` are not `Number` subtypes, so juliacall neither converts a Python callable/sequence into them nor overloads their operators automatically. Per #76, the fix lives as a PythonCall extension in the package where the functionality is defined. Operator overloading on the Python side is handled by a thin wrapper forwarding onto the operators these Julia types already define; this PR only needs to make construction work.

## Scope / notes

- No change to existing code paths; the new methods dispatch only on `::Py`.
- A Python callable is invoked per evaluation, which requires the GIL and so is CPU-only. This is inherent to any Python callback and matches how the `scibmad` package behaves today.
- `BatchParam` treats a Python object with `__len__` as a sequence, otherwise a scalar (degenerate) batch.

## Testing

Verified via juliacall (with `BeamTracking` + `PythonCall` dev-loaded, extension auto-loading confirmed) that `TimeDependentParam(lambda t: 2*t+1)(3.0) == 7.0`, the constant form, and `BatchParam([0.1, 0.2, 0.3])` all work. Opened as a **draft** since I could not run the full BeamTracking.jl CI locally; a `PythonCall` test target could be added if wanted.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -64,8 +64,9 @@ Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 GTPSA = "b27dd330-f138-47c5-815b-40db9dd9b6e8"
 JET = "c3a54625-cd67-489e-a8e7-0a5a0ff4e31b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+PythonCall = "6099a3de-0909-46bc-b1f4-468b9a2dfc0d"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "Distributions", "JET", "GTPSA", "BenchmarkTools", "Beamlines", "StaticArrays", "LinearAlgebra"]
+test = ["Test", "Distributions", "JET", "GTPSA", "BenchmarkTools", "Beamlines", "StaticArrays", "LinearAlgebra", "PythonCall"]
--- a/Project.toml
+++ b/Project.toml
@@ -27,9 +27,11 @@ Unrolled = "9602ed7d-8fef-5bc8-8597-8f21381861e8"
 [weakdeps]
 Beamlines = "5bb90b03-0719-46b8-8ce4-1ef3afd3cd4b"
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+PythonCall = "6099a3de-0909-46bc-b1f4-468b9a2dfc0d"
 
 [extensions]
 BeamTrackingBeamlinesExt = "Beamlines"
+BeamTrackingPythonCallExt = "PythonCall"
 
 [compat]
 Accessors = "0.1.42"
@@ -44,6 +46,7 @@ HostCPUFeatures = "0.1"
 KernelAbstractions = "0.9.35"
 LinearAlgebra = "1"
 MacroTools = "0.5.16"
+PythonCall = "0.9"
 ReferenceFrameRotations = "3"
 SIMD = "3.7.1"
 SIMDMathFunctions = "0.1.3"
--- a/ext/BeamTrackingPythonCallExt/BeamTrackingPythonCallExt.jl
+++ b/ext/BeamTrackingPythonCallExt/BeamTrackingPythonCallExt.jl
@@ -1,0 +1,43 @@
+module BeamTrackingPythonCallExt
+using PythonCall
+import BeamTracking: TimeDependentParam, BatchParam
+
+# Construction of BeamTracking's deferred-parameter types from Python objects.
+#
+# `TimeDependentParam` and `BatchParam` are not `Number` subtypes, so juliacall
+# neither overloads their operators nor converts Python callables/sequences into
+# them automatically. These methods let a Python callable become a
+# time-dependent parameter, and a Python sequence become a batch parameter,
+# converting values to `T` (default `Float64`) at evaluation time.
+#
+# Operator overloading on the *Python* side (so `2 * Time() + 1` works from
+# Python) is provided by the operators these Julia types already define; a Python
+# wrapper only needs to forward `+ - * / ^` onto them.
+
+_pycallable(f::Py) = pytruth(pybuiltins.callable(f))
+
+# --- TimeDependentParam ------------------------------------------------------
+
+function TimeDependentParam(f::Py, ::Type{T}) where {T}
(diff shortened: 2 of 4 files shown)


DavidSagan:
Added test coverage for the new extension (3d69487): `PythonCall` wired into the test target + a `BeamTrackingPythonCallExt` testset covering `TimeDependentParam(::Py)` (callable + constant) and `BatchParam(::Py)` (sequence + scalar). Same rationale as the Beamlines PR — the extension had 0% coverage because the suite never loaded PythonCall, and PySciBmad's tests are a separate repo/Codecov project. Note `using PythonCall` provisions Python via CondaPkg in CI. Verified locally (4/4).

DavidSagan:
The failing check (**Julia lts - windows-latest - x64**) is a pre-existing failure on `main`, unrelated to this PR.

The job dies with a Julia/LLVM backend crash — not a test assertion — while compiling the IBS tracking kernel:

```
LLVM ERROR: Do not know how to split the result of this operator!
in expression starting at test\BeamlinesExt\beamlines_ibs_test.jl:3
  #universal!#6 at ext\BeamTrackingBeamlinesExt\unpack_bl.jl:289
```

Evidence this is not from this PR:

1. The `BeamTrackingPythonCallExt` testset added here **passes** on that same lts-Windows job (`✓ BeamTracking → BeamTrackingPythonCallExt`) — the crash happens later, in the untouched `beamlines_ibs_test.jl`.
2. This PR only touches `Project.toml`, `ext/BeamTrackingPythonCallExt/…`, `test/PythonCallExt_test.jl`, and `test/runtests.jl`. The crash is in `unpack_bl.jl` / `ibs_kick.jl`, all unchanged.
3. The most recent CI run on `main` (merge of #313) fails on the identical job with the identical `LLVM ERROR: Do not know how to split the result of this operator!` at `beamlines_ibs_test.jl:3`.

So this is an upstream LLVM 15 codegen limitation on the Julia 1.10 (LTS) + Windows x64 combination, triggered by the vectorized IBS kernel — independent of the PythonCall extension work here. All other 11 jobs (including lts on Linux/macOS and Windows on 1.11/1.12) pass.

DavidSagan:
*(Claude, via @DavidSagan.)*

@mattsignorelli — you're right, and here more strongly than on Beamlines.jl#158: most of this extension shouldn't exist. Ran the cases against BeamTracking 0.7.1 + PythonCall rather than reasoning about them.

### `BatchParam(::Py)` — delete, no replacement needed

`BatchParam(::AbstractArray)` already accepts a Python array today, zero-copy:

```
pyconvert(Any, np.array([0.1,0.2,0.3])) -> PyArray{Float64, 1, true, true, Float64}   <: AbstractArray
BatchParam(that)                        -> batch::PyArray{Float64,…}, eltype Float64  ✅
```

So `BatchParam(numpy_array)` works with no extension and no copy — and a real typed array is what you want in a batched/GPU run anyway. The only input that doesn't already work is a plain Python `list`: it converts to `PyList{Any}`, which *is* an `AbstractArray` and *is* accepted, but yields an `Any`-eltype batch, which is the wrong thing to feed tracking. Turning a list into a typed array is a Python-side concern (one `pyconvert(Vector{Float64}, ·)`, which PySciBmad already does), not something BeamTracking should carry methods for.

### `TimeDependentParam(::Py)` — the real fix is a type generalization in `src/time.jl`

This isn't a Python problem. Both of these fail identically today:

```julia
TimeDependentParam(py_lambda, false)   # MethodError: no method matching TimeDependentParam(::Py, ::Bool)
TimeDependentParam(Doubler(), false)   # MethodError: no method matching TimeDependentParam(::Doubler, ::Bool)
```

where `Doubler` is an ordinary Julia callable struct. The cause is `src/time.jl`:

```julia
struct TimeFunction{F<:Function}                       # -> {F}
  f::F
end
TimeDependentParam(f::Function, _isconst) = ...        # -> f, untyped
```

`Function` excludes every callable that isn't a subtype of it — functors, `FunctionWrapper`s, and `Py`. Relaxing those two annotations is the "type generalization" route bmad-sim/SciBmad.jl#76 asks for, fixes the Julia-side cases at the same time, and needs no `PythonCall` weakdep, no `[extensions]` entry, no new module.

One genuinely Python-only residue: a `Py` call returns a `Py`, measured —

```
typeof(py_lambda(3.0)) == Py        # not Float64
```

— so *something* has to `pyconvert` the result, or the parameter hands `Py` values into the kernels. That's a single wrapping closure and it belongs on the Python side, where PySciBmad already has it (`t -> pyconvert(T, pyf(t))` in `_glue.jl`). With the type generalization upstream, that stays as three lines downstream and BeamTracking stays Python-free.

### Also worth flagging independent of size

`TimeDependentParam(f::Py)` (single argument) invents an API that doesn't exist in Julia — there is no single-argument function constructor at all:

```
TimeDependentParam(f::TimeFunction, _isconst)
TimeDependentParam(f::Function, _isconst)
TimeDependentParam(a::Number)
TimeDependentParam(a::TimeDependentParam)
```

Julia users must pass `_isconst`; the extension would let Python users omit it. Divergent constructor conventions between the two front ends is a thing to avoid on its own.

### Proposal

Close this PR, and instead take a 2-line type generalization in `src/time.jl` (`TimeFunction{F}`, untyped `f`) as its own PR — happy to open it. That also drops `PythonCall` from `[targets] test`, so the CondaPkg bootstrap comes back out of CI here.

The pre-existing lts-Windows CI failure noted above (`LLVM ERROR: Do not know how to split the result of this operator!` in `beamlines_ibs_test.jl`, reproducing on `main`) is unrelated to any of this and still worth its own issue.

mattsignorelli:
Do the proposal and submit a PR.....

PR #321, opened 2026-08-04
Cut CI wall time by sharding tests and instrumenting one job

DavidSagan:
CI test jobs have been running 2–4 hours — [this one][job] took 2h41m, and the
Julia 1.11 / Windows job on the same run took 4h6m — and `timeout-minutes` has
been raised twice to keep them from being killed. This cuts the critical path
without removing, weakening, or skipping a single assertion.

[job]: https://github.com/bmad-sim/BeamTracking.jl/actions/runs/30913748571/job/92006559603

## Where the time goes

Roughly **94% of the suite is JIT compilation**, not tracking. Running
`time_test.jl` twice in the same session:

```
pass 1: 48.0 s     pass 2: 3.0 s
```

Nearly every test builds a distinct `KernelChain` tuple type, and
`__generic_kernel!` is `@generated` + `@unroll`-expanded over it, so each one
recompiles a large fused kernel from scratch. That makes the suite very
sensitive to anything that inflates generated IR — which is what CI was doing,
on all twelve matrix jobs at once. Measured on `time_test.jl`:

| configuration | time | vs. plain |
|---|---|---|
| plain | 49.7 s | 1.00x |
| `--code-coverage` | 65.9 s | 1.33x |
| `--check-bounds=yes` | 111.9 s | 2.25x |
| both (what CI ran) | 149.4 s | **3.0x** |

`julia-actions/julia-runtest` defaults to `coverage: true` and
`check_bounds: yes`, so every job paid that 3x.

The other half of the problem is shape: one testset, `Beamlines`, was 58% of
the job (92m53s of 161 min), so the suite had a serial floor that no amount of
matrix parallelism could lower.

## Changes

**Instrument one platform instead of twelve.** `coverage` and
`check_bounds: yes` now run only on `lts` + `ubuntu-latest`. Codecov merges the
per-group uploads into a single report, so coverage reporting is unchanged, and
bounds checking still runs over every test once per commit rather than twelve
times.

**Shard the suite into three groups** selected by `BEAMTRACKING_TEST_GROUP`,
read by `test/runtests.jl`. With the variable unset — a plain `Pkg.test()` —
the whole suite runs exactly as before. This required moving the `Yoshida`
testset out of `BeamlinesExt_test.jl` into `BeamlinesExt_symplectic_test.jl`;
the moved block is byte-for-byte identical, and `include("lattices/esr.jl")`
moves with it since `ring` is used only there.

`runtests.jl` also gained a guard that errors if any `*_test.jl` file in
`test/` belongs to no group, so a new test file cannot silently stop running.

**Every os/arch runs on every event, on `lts` and release.** All four os/arch
combinations from `main` are kept; the version axis drops `1.11`, so the
matrix is 8 platforms rather than 12 (see the revision notes below). The
release row is spelled `'1'` so it follows the current stable series without
an edit when 1.13 ships. The matrix also moved from `version`/`os`/`arch` axes
with `exclude`/`include` into a single explicit `platform` list, because
`include:` entries conflict on all three keys and GitHub would add them as
standalone combinations with no `group` key. `timeout-minutes` drops from 180
to 60, which is now a meaningful ceiling rather than one that gets raised
whenever it is hit.

## Verification

Full suite run locally before and after, comparing per-testset pass counts, plus
each group run standalone and a `Pkg.test()` check that the environment variable
reaches the test process the way `julia-runtest` will set it.

| testset | baseline (`main`) | after (all groups) |
|---|---|---|
| Miscellaneous | 1 | 1 |
| SaganCavityKernel | 15 | 15 |
| Beamlines | 186 | 69 + 117 |
| AlignmentKernel | 16 | 16 |
| ApertureKernel | 8 | 8 |
| BeamTracking | 210 | 210 |
| Yoshida | 45 | 45 |
| IBS utilities | 22 | 22 |
| Callbacks | 22 | 22 |
| Implicit | 28 | 28 |
| Batch | 156 | 156 |
| Time | 86 | 86 |
| **total** | **773** | **773** |

All four runs exit 0: each group standalone (`symplectic`, `batch_time`, `core`)
and the default all-groups run. A `Pkg.test()` invocation confirms
`BEAMTRACKING_TEST_GROUP` reaches the test process the way `julia-runtest` sets
it.

## Out of scope, but worth separate issues

Both of these affect users, not just CI:

- **Julia 1.11/1.12 is 3.7–5x slower than lts on this code.** Same PR, same
  runner class, identical test counts: `Beamlines` 18m32s on lts vs 92m53s on
  1.11, `Batch` 5m09s vs 20m38s, `Time` 7m08s vs 26m31s. Since it is compile
  time, it is also first-call latency for anyone on a current Julia.
- **`--optimize=1` cuts compile-bound test time ~2.1x** (69.7 s vs 149.4 s under
  the CI flags). Not enabled here: it would change the `@ballocated(...) == 0`
  results in `test_matrix`/`test_map`, which deserves a deliberate trial rather
  than being bundled into this change.

## Measured result (first CI run on this branch)

Same commit lineage, old config on `main` vs. this branch:

| platform | before (one job) | after (3 shards, critical path) | speedup |
|---|---|---|---|
| lts / ubuntu (instrumented) | 22.4 min | 10.8 min | 2.1x |
| 1.12 / ubuntu | 36.3 min | 9.0 min | 4.0x |
| 1.12 / windows | 45.9 min | 11.9 min | 3.9x |

Whole run: **13 min** end to end, vs. 140 min for the same commit on `main`.

The two rows separate the two mechanisms cleanly. lts/ubuntu is the instrumented
job, so it still pays coverage + `check_bounds` in full and its 2.1x is sharding
alone; the 1.12 jobs get both and land near 4x.

That run predates the revision below, so it covers 3 of the 12 platforms. Note
also that this is `main`'s code — #319's integrator changes multiply whatever
these jobs cost, so that branch will still be slower than the figures above.

## Revision: full matrix restored on pull requests (superseded by the next section)

Per @JosephPeterDevlin's review — the reduced PR matrix was the one part of this
that traded coverage for time, and sharding makes the trade unnecessary. All 12
platforms now run on every event, as before; the weekly cron that backstopped
the subset is gone with it.

This is affordable because sharding's 3x setup overhead is more than paid for
by the instrumentation savings. Measured, full 12-platform matrix on this
commit against the same 12 platforms on `main`:

| | `main` (run 30582013647) | this branch (run 30969424491) |
|---|---|---|
| jobs | 12 | 36 |
| total job-minutes | 497.9 | **295.1** |
| longest single job | 68.4 min | **11.4 min** |
| wall clock | 140 min | **21 min** |

So the full matrix now costs ~41% *less* machine time than it does on `main`,
not more. Only the instrumented platform got more expensive per-platform, since
it is the one still paying the 3.0x for coverage and `check_bounds`.

Concurrency behaves as expected: GitHub started 19 of the 36 jobs immediately,
so the matrix runs in two waves and still finishes in 21 minutes.

## Revision: `lts` and release only

Per @JosephPeterDevlin's follow-up: the version axis drops `1.11`, leaving
`lts` and the latest stable release. All four os/arch combinations are
unchanged and each runs on both versions, so @mattsignorelli's requirements
still hold — no test removed, every architecture/OS retained, every one of them
on `lts` and release. Release is spelled `'1'` rather than `'1.12'` so it tracks
the current stable series on its own.

Matrix: 8 platforms × 3 groups = 24 jobs, down from 36. Measured, all 24 green
([run 31029853858](https://github.com/bmad-sim/BeamTracking.jl/actions/runs/31029853858)):

| | `main` (12 jobs) | 12 platforms (36 jobs) | 8 platforms (24 jobs) |
|---|---|---|---|
| total job-minutes | 497.9 | 295.1 | **200.3** |
| longest single job | 68.4 min | 11.4 min | **13.2 min** |
| wall clock | 140 min | 21 min | **13.4 min** |

GitHub started 20 of the 24 jobs immediately, so this is effectively one wave
against the ~20-job ceiling instead of the two the 36-job matrix needed — which
is where the wall clock drop from 21 to 13.4 minutes comes from. `'1'` resolved
to Julia 1.12.6 on this run.

## A pre-existing flake this surfaced (not caused by this PR)

The first full-matrix run had one job fail with:

```
LLVM ERROR: Do not know how to split the result of this operator!
in expression starting at test/BeamlinesExt/beamlines_ibs_test.jl:3
```

This is already on `main`. Jobs 87442179068 (2026-07-15) and 87614296182
(2026-07-16), both `lts - windows-latest`, failed with the same LLVM error,
same file, same line, each with the same 17 CPU-feature warnings — 2 of the
last 15 `main` runs.

The trigger is `julia-actions/cache` restoring a Julia depot precompiled on a
runner with different CPU features. The correlation is exact:

| | `Will continue with incorrect CPU feature flag: -avx512*` | result |
|---|---|---|
| run 30960126327, same job | 0 warnings | pass |
| run 30969424491 attempt 1 | 17 warnings | **LLVM crash** |
| attempt 2 (restored attempt 1's cache) | 17 warnings | **LLVM crash** |
| attempt 3, after deleting that cache entry | 0 warnings | pass |

Deleting the one poisoned cache entry made it pass on a cold depot, which is
why this run is green. Note that attempt 2 is not an independent trial: a rerun
restores the same cache (`run_attempt=1` appears in the restored key), so it
reproduces deterministically until the entry is evicted.

Worth fixing separately, most likely by pinning `JULIA_CPU_TARGET` so cached
pkgimages are portable across runner microarchitectures. Deliberately not
bundled here: it changes codegen, and the suite has `@ballocated(...) == 0`
assertions in `test_matrix`/`test_map` that are sensitive to that.

One caveat this PR does own: exposure scales with job count. At 24 jobs instead
of 12 there are ~2x the cache restores per run, so this flake will surface more
often until it is fixed. (Shards do not pollute each other — the cache key
includes `group=`, so each shard has its own entry.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DavidSagan:
@JosephPeterDevlin @mattsignorelli Please review.

JosephPeterDevlin:
Splitting up the groups and only running `--check-bounds` on LTS is smart (same for `--code-coverage` but I'm not sure this makes much of a difference). The only thing I'm unsure about is reducing the number of architectures/versions. There is a maximum number of architectures we can run at once (I think 20 total jobs?), but we should be able to increase this.

DavidSagan:
@JosephPeterDevlin Good catch — that was the one part of this that traded coverage for time, and it turns out sharding makes the trade unnecessary. Pushed 2769b83: **all twelve platforms run on every event again**, same set as `main`, and the weekly cron that backstopped the subset is gone with it.

The reason it's affordable is that sharding's 3x setup overhead is roughly paid for by the instrumentation savings you mentioned. Summing the three shards against the single job they replaced:

| platform | before (1 job) | after (3 shards, sum) | after (longest shard) |
|---|---|---|---|
| lts / ubuntu (instrumented) | 23.1 min | 27.0 min | 11.6 min |
| 1.12 / ubuntu | 37.2 min | 28.3 min | 9.6 min |
| 1.12 / windows | 48.8 min | 34.8 min | 12.9 min |

The instrumented platform is the only one that costs more, since it's the one still paying the 3.0x — so `--code-coverage` plus `--check-bounds` does make a large difference, it's just concentrated in one job now. The other eleven run ~25% cheaper than today. Whole matrix lands around 400 job-minutes vs. the 498 that [run 30582013647](https://github.com/bmad-sim/BeamTracking.jl/actions/runs/30582013647) spent on `main`.

What does go up is job count, 12 → 36, which is exactly the concurrency ceiling you're pointing at. 36 against a 20-job limit is two waves, and 9 of the 36 are macOS against a 5-job macOS limit. Even fully serialized into waves that's well inside the old 140-minute wall clock, and the longest single job drops from 68 min to ~13. The CI run on this commit is the first full-matrix run under the new config, so it measures the queueing directly instead of relying on my arithmetic — worth looking at before merge.

One structural note: the platform set is identical to `main`'s, but it's now one explicit `platform:` axis instead of `version`/`os`/`arch` with `exclude`/`include`. That's forced — the `include:` entries conflict on all three keys, so GitHub adds them as standalone combinations, which would carry no `group` key and run with `BEAMTRACKING_TEST_GROUP` unset.

DavidSagan:
Full matrix is green: [run 30969424491](https://github.com/bmad-sim/BeamTracking.jl/actions/runs/30969424491), 36 jobs, all 12 platforms.

It came in better than my estimate — restoring the full matrix costs **less** machine time than `main` does today, not more:

| | `main` (run 30582013647) | this branch |
|---|---|---|
| jobs | 12 | 36 |
| total job-minutes | 498 | **287** |
| longest single job | 68.4 min | **11.4 min** |
| wall clock | 140 min | **21 min** |

On your concurrency point: GitHub started 19 of the 36 immediately, so the ceiling is right where you guessed (~20). It runs in two waves and still finishes in 21 min, so I don't think it needs raising.

---

**One thing worth your attention, @JosephPeterDevlin, since you've been in this file before.** The first attempt had a job fail with:

```
LLVM ERROR: Do not know how to split the result of this operator!
in expression starting at test/BeamlinesExt/beamlines_ibs_test.jl:3
```

Not caused by this PR — it's already on `main`. Jobs 87442179068 (7/15) and 87614296182 (7/16) failed with the same error, same file, same line. Same neighborhood as your `9de6242 "Fix SIMD bug"`.

The trigger is `julia-actions/cache` restoring a Julia depot precompiled on a runner with different CPU features. The correlation is exact:

| | `Will continue with incorrect CPU feature flag: -avx512*` | result |
|---|---|---|
| previous run, same job | 0 warnings | pass |
| attempt 1 | 17 warnings | **LLVM crash** |
| attempt 2 (restores attempt 1's cache) | 17 warnings | **LLVM crash** |
| attempt 3, after deleting that cache entry | 0 warnings | pass |

Deleting the poisoned entry fixed it, which is why the run is green. Worth knowing that a plain re-run does *not* clear this — the rerun restores the same cache (`run_attempt=1` shows up in the restored key), so it reproduces deterministically until the entry is evicted. That probably explains any past "re-ran it and it still failed" experiences with this test.

The likely fix is pinning `JULIA_CPU_TARGET` so cached pkgimages are portable across runner microarchitectures. I deliberately did **not** bundle it here: it changes codegen, and `test_matrix`/`test_map` have `@ballocated(...) == 0` assertions that are sensitive to that. Happy to do it in this PR or a separate one — your call.

The caveat this PR does own: at 36 jobs instead of 12 there are ~3x the cache restores per run, so this flake will show up more often until it's fixed. (Shards don't pollute each other — the cache key includes `group=`.)

mattsignorelli:
My only requirements are that 
1. Zero tests has been removed
2. All current architectures/OSs are retained in the test matrix
3. All architectures/OSs are tested on lts and release

JosephPeterDevlin:
> My only requirements are that
> 
> 1. Zero tests has been removed
> 2. All current architectures/OSs are retained in the test matrix
> 3. All architectures/OSs are tested on lts and release

These seem to all be met. We should be able to make it more efficient by removing 1.11 tests. @DavidSagan Can you make it so that tests are only run on LTS and the latest stable release?

DavidSagan:
@JosephPeterDevlin Done — pushed 5cba11c. The version axis is now `lts` and the
latest stable release; `1.11` is gone.

One spelling choice worth flagging: the release row is `'1'`, not `'1.12'`, so
`setup-julia` resolves it to whatever the current stable series is and the file
does not need editing when 1.13 ships.

@mattsignorelli your three requirements still hold: no test removed (773 → 773,
table in the description), every architecture/OS retained, and each of them now
runs on both `lts` and release rather than a subset.

Matrix is 8 platforms × 3 groups = 24 jobs, from 36:

| | `main` (12 jobs) | previous revision (36 jobs) | now (24 jobs) |
|---|---|---|---|
| total job-minutes | 497.9 | 295.1 | **194.3** |
| longest single job | 68.4 min | 11.4 min | **12.0 min** |

The last column is [run 30969424491](https://github.com/bmad-sim/BeamTracking.jl/actions/runs/30969424491)
with its twelve 1.11 jobs subtracted — the 1.11 platforms were 100.8 of that
run's 295.1 job-minutes. 24 jobs also sits much closer to the ~20-job
concurrency ceiling you raised, so it should be roughly one wave instead of two.

CI on the new matrix is running now ([run 31029853858](https://github.com/bmad-sim/BeamTracking.jl/actions/runs/31029853858)); I will post the measured numbers when it finishes.

DavidSagan:
Measured, and it came in slightly better than the projection I posted above.
[Run 31029853858](https://github.com/bmad-sim/BeamTracking.jl/actions/runs/31029853858):
all 24 jobs green.

| | `main` (12 jobs) | previous revision (36 jobs) | now (24 jobs) |
|---|---|---|---|
| total job-minutes | 497.9 | 295.1 | **200.3** |
| longest single job | 68.4 min | 11.4 min | **13.2 min** |
| wall clock | 140 min | 21 min | **13.4 min** |

(200.3 vs. the 194.3 I projected — the subtraction was close, the difference is
per-job runner variance.)

@JosephPeterDevlin the concurrency point resolves cleanly: GitHub started 20 of
the 24 immediately, so it is one wave instead of the two the 36-job matrix
needed, and that is where the wall clock drop from 21 to 13.4 min comes from.
No need to raise any limit.

`'1'` resolved to Julia 1.12.6 on this run, so the release coverage is identical
to what `'1.12'` gave, without the file needing an edit later.

JosephPeterDevlin (approved):

PR #323, opened 2026-08-10
Newcallback

mattsignorelli:


mattsignorelli:
tested on GPU and everything is consistent

Note that with `ramp_update_each_particle=false`, a ramp is only done at the START of the element, so beta_gamma_enter = beta_gamma_exit.

With  `ramp_update_each_particle=true`, through the entire element particles are ramped to their single-particle reference energies and tracked, then at the end of the element they are ramped singularly to the reference energy at the reference time at the END of the element

In practice t_initial \approx t_exit so this difference is completely negligible but important to keep in mind....

things changing significantly with reference energy inside element would require extra work

PR #333, opened 2026-08-27
Absolute time tracking

eiad-hamwi:
This is a minimal extension to RF tracking to take `bunch.t_ref` into account when calculating RF phase.

Adds an `absolute_time_tracking` flag for `track!` which is propagated downstream.

This is needed in SciBmad's `find_closed_orbit` and `twiss` so that "radial steering" is possible by manipulating RF frequency.


mattsignorelli:
Relevant discussion: https://github.com/bmad-sim/BeamTracking.jl/issues/213

eiad-hamwi:
It sounds like `absolute_time` makes more sense as a key in `RFParams` rather than a flag in `track!`

